Issue #3, opened 2021-12-18
Implement var- and obs- aligned multidimensional arrays (obsm, varm)


ambrosejcarr:
Additionally, explore whether there are opportunities to specify metadata standards for derived analysis results (e.g. reduced dimensionality representations)

From @LTLA

> Suggest developing some metadata standards on the auxiliaries so that downstream tools can meaningfully use them. Otherwise e.g. if someone sees an array in there they need to ask "is this reducedDims? Or gene signature scores? Or FACS data?" etc. We have several visualization tools that poke around in the SCE's reducedDims to figure out what to plot.

From "open questions" in the gdoc version:

> Auxiliary arrays need more thought as they are commonly indexed by one of "var" or "obs" labels.  Do we need to have the API enforce this (probably depends on whether or not there are features dependent on it). Use case: if UMAP embedding is stored as an aux array, it wants to be indexed by same labels as "obs"

ambrosejcarr:
### Problem

I think there is agreement that`obsm` and `varm` are needed to store multidimensional annotations. 

Multidimensional annotations may be created from an individual `sc_group` or via multiplex methods that leverage more than one group. Most examples (e.g. Figure B in the [Muon publication](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-021-02577-8)) use two modalities, but poly-modal assays will arise, so I recommend we use examples of with three modalities to drive decision making. 

Consider an experiment with three data modalities: 

```
sc_dataset
- sc_group (RNA)
- sc_group (ATAC)
- sc_group (Protein)
```

A common analysis that stores results in `obsm` is the estimation of a dimension-reduced latent space. Methods can estimate latent spaces based on one or more modalities. Here, there are several options: 

```
- RNA
- ATAC
- Protein
- RNA + ATAC
- RNA + Protein
- ATAC + Protein
- RNA + ATAC + Protein
```

I see two requirements:

1. Each type of latent space above must be storable
2. Users must be able to understand how the latent spaces should be used, and how it should not be used.

### Possible Solutions

I see two, but am interested if clever individuals can do better.

#### 1. Create `obsm` and `varm` slots for `sc_dataset` and `sc_group`. 

`obsm` associated with `sc_groups` can be inferred to have been built exclusively from data within that group, but no guarantees can be made (i.e. the format will not validate that additional information has not been included). `obsm` associated with the `sc_dataset can be inferred to have been built from more than one group, but again, this will not be validated. Missing values are permitted. 

[...]

PR #17, opened 2022-03-05
Suggestion re introducing sc_group vs sc_dataset

falexwolf:
Extending discussion here: https://github.com/single-cell-data/matrix-api/issues/11

* This would change the sequence and first introduce `sc_group` & `sc_dataset` and then talk about the specification within `sc_group`.
* It would make identifications with existing formats on the level of `sc_group` and not on the level of the fields within `sc_group`.

There is also a question here regarding "allowing arbitrary collections of dataframes" vs. aligned multi-dimensional annotations (https://github.com/single-cell-data/matrix-api/issues/3).

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -36,18 +36,18 @@ It first describes a general data model that captures all the above frameworks. 
 
 ## The Data Model
 
-Based on our research and avoiding a lot of the jargon, there are the following “components” in all the single-cell projects:
+Based on our research and avoiding a lot of the jargon, there are the following “components” in typical single-cell projects, which account for various means of representing filtered & transformed datasets, and associating data from different assays/modalities.
 
-* One or more **labeled 2D arrays** that we will call `X` (similar to AnnData and CXG - Seurat and SingleCellExperiment call them `assays`, and Loom calls them `layers`).
+* A **sc_group**, a characterization of a set of cells that can be cast into a matrix structure with extensive annotations. An `sc_group` often groups variables of the same type and assay (say, `RNA`, `ATAC`, `Protein`) and cells sourced from comparable homogeneously processed samples. An `sc_group` would be called `AnnData` in anndata, and `assay` in `SingleCellExperiment` and Seurat. Every `sc_group` exists within a `sc_dataset`.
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -36,18 +36,18 @@ It first describes a general data model that captures all the above frameworks.
...
+Within an `sc_group`, we find
 
-* A **sc_group**, composed of the above elements -- a self-consistent set of `X` matrices (aka _layers_) and their axis labels (`obs`,`var`), where all three share consistent shape and indexing labels.  Every `sc_group` exists within a `sc_dataset`.
-* A **sc_dataset**, which is a set of named `sc_groups`.  Primary use is representing a collection of related annotated matrix data, including progressively filtered data (raw vs norm), related assays with different labels/shape, etc.
+* One or more **labeled 2D arrays** that we will call `X` (as within AnnData and CXG - Seurat and SingleCellExperiment call them `assays`, and Loom calls them `layers`).
+* A **dataframe** called `obs`, which contains the row labels for the `X` arrays (in SingleCellExperient the `X` arrays are transposed, so `obs` are the column labels).
+* A **dataframe** called `var`, which contains the column labels for the `X` arrays (in SingleCellExperient the `X` arrays are transposed, so `var` are the row labels).
+* Any **key-value metadata** called `uns`.
+* Any other number of auxiliary **dataframes** and **arrays**. [Why do we think we don't need `obsm` and `varm`?]
```
ambrosejcarr, line 50:
```suggestion
* Any other number of auxiliary **dataframes** and **arrays**.
```

We do need obsm and varm. I suggest we address that in a separate PR, which would unblock merging of this one.

I've update #3 with some thoughts and questions for you.

ambrosejcarr (approved):
These changes look good and I'd be happy to see them merged.

@aaronwolen Could you provide a second opinion? Is there anything here you think should be changed?

aaronwolen:
Looks great. I agree that starting with descriptions of the higher-level components and including their anndata/seurat/bioc analogs will help readers get oriented more quickly. 

Related but probably for a separate PR:
- I think we should bump-up the `matrix.png` diagram to come before *Notes:*
- and some of those bullets could be folded into the revised descriptions (e.g., *There can be more than one `X` arrays...*)

ambrosejcarr:
Thanks for this suggestion @falexwolf !

falexwolf:
Thanks for merging! 😄

PR #20, opened 2022-03-09
Consolidate data model description and notes

### LINKED ISSUE #19: Condense specification summary and matrix.png description
@aaronwolen suggests:

> * I think we should bump-up the matrix.png diagram to come before Notes:
> * and some of those bullets could be folded into the revised descriptions (e.g., There can be more than one X arrays...)

aaronwolen:
Relevant bullets from *Notes:* have been folded into the appropriate data model description. One exception: the note about implementation-specific details is now a footnote (but we can do something different if there's opposition to the use of footnotes). 

This moves the `sc_group` diagram closer to its description and closes #19.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -43,20 +43,11 @@ Based on our research and avoiding a lot of the jargon, there are the following 
 
 Within an `sc_group`, we find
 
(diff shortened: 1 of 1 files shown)


ambrosejcarr (approved):
Looks good to me. Sorry for the slow review.

Issue #18, opened 2022-03-07
Data types


ivirshup:
What data types have to be supported here? And should this be extensible?

By data type I’m thinking of programming types, like string, int, etc. An initial set of types:

### Collection types

* n-dimensional array
* nullable array
* sparse array (csc, csr, coo)
* tables (dataframe, collection of arrays)
* ragged arrays

### Data types

* string (exactly how is complicated)
* categorical
* 8, 16, 32, 64 bit float, int, and uint
* nullable int, float, bool
* datetime?

A very easy way to go here would to just take an existing type system – like Arrow’s.

evanbiederstedt:
> And should this be extensible?

What do you mean by "extensible" here? @is

> What data types have to be supported here? 
> By data type I’m thinking of programming types, like string, int, etc. An initial set of types:

In terms of fundamental data types, I think this is another argument for working on the C++ API actually, cf https://github.com/single-cell-data/matrix-api/issues/2. 

Python + R + JS don't have the same fundamental data types, but encoding these within something like C++ first allows for more clear translations. 

In terms of collections:

I ask myself how much correspondence there is within the current single-cell frameworks. Consider something like the compressed sparse matrix: I'm guessing the Python packages rely on the SciPy implementation or equivalent: https://github.com/single-cell-data/matrix-api/blob/main/specification.md#proposed-single-cell-python-api

R packages use `dcgMatrix` from `Matrix`: https://stat.ethz.ch/R-manual/R-devel/library/Matrix/html/dgCMatrix-class.html

So for collections, how different is the "real implementation" for these collections written in C/C++? Or do we just accept that R/Python/JS are different beasts and use the existing frameworks (an argument against using the C/C++ API for the resulting R/Python/JS libraries)?

This might be worth a deep dive.

pablo-gar:
With the exception of nullable array and ragged arrays, everything has been implemented in `tiledbsoma`. There are no plans in the foreseeable future to dedicate resources to it.

PR #32, opened 2022-06-26
design discussions

bkmartinjr:
PR to track discussion around some design discussions about base level primitives and typing


brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+In this document, the term `dataframe` implies something akin to an Arrow `RecordBatch`, R `data.table` or Python `pandas.DataFrame`, where:
+
+- dimensions are string-indexed
+- multiple columns may exist, each with a string column name
+- all columns are individually typed, monomorphic, and contain simple types (eg, int64)
+- all columns are of equal length
+
+All SOMA data objects are named with URIs.
+
+> ⚠️ **Issue** -- this section would benefit from a data model diagram, helpfing visualize the relationships between the types.
```
johnkerl, line 46: +1

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+> ⚠️ **Issue** -- the use of `SOMA` to name the API _and_ a data type is confusing. We should consider a new name for the aforementioned SOMA _type_. _Candidate:_ `SOMAAnnotatedMatrix`?
```
johnkerl, line 48: really good call -- this is another 'high-level issue' we need to deal with to avoid uptake friction

i like:

* keep SOMA API for the API
* keep for the (current) object, rename it away from `SOMA` to something else -- not sure what -- open to ideas

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+In the following doc:
+
+- `primitive` types in this specification refers to Arrow primitive types, eg, `int32`, `float`, etc.
+- `string` refers to Arrow UTF-8 variable-length `string`, ie, `List<Char>`.
+- `simple` types include all primitive types, plus `string`.
+
+Other Arrow types are explicitly noted as such, eg, `Arrow.RecordBatch`.
+
+> ⚠️ **Issue** - is the use of Arrow as a base type language acceptable? There are others, up to and including inventing a new system, but it seemed prudent to adopt an existing, language neutral specification, which is already supported by many "engines".
```
johnkerl, line 64: i've no strong feels here & happy to defer to others' wisdoms

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+> ⚠️ **Issue** - are there parts of the Arrow type system that we wish to _explicitly exclude_ from SOMA? I have left this issue open (ie, no specific text) for now, thinking that we can come back and subset as we understand what complex types are required, and how much flexibility should be in this spec. We clearly need some complex types (eg, RecordBatch, List, etc) as they are implied by `string`, etc. My own preference would be to mandate a small set of primitive types, and leave the rest open to implementations to support as they feel useful.
+
+## Metadata (uns)
+
+All SOMA objects may be annotated with a small amount of metadata, in the form of a simple key/value map:
+
+- `uns` (map[string, simple]) - string-keyed mapping to any **simple** type (eg, int32, string, etc).
+
+Only `simple` types are supported in the metadata values. The metadata lifecycle is the same as its containing object, eg, it will be deleted when the containing object is deleted.
```
johnkerl, line 74: we need to say what 'simple' is. allowing a metadata value to be a list of strings seems really crucial -- is a list of strings 'simple'?

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+> ⚠️ **Issue** - `uns` - do we need more complex typing for uns values, and if so, how to represent capture what can is required to be supported? (eg, anything Arrow can represent? a subset?)
+>
+> **Proposal**: restrict `uns` to a mapping of string keys and `simple` values. Extend to complex types in the future as we have concrete use cases which warrant the complexity.
+
+> ⚠️ **Issue** - the name `uns` is not really all that suggestive. **Proposal:** rename `uns` to `metadata`, eg, `print(soma.metadata['version'])`
```
johnkerl, line 80: +1 -- this way neither seurat `misc` or anndata `uns` 'wins'; we have `neutral territory` in our naming

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+## SOMACollection
+
+SOMACollection is an unordered, `string`-keyed map of values of type SOMADataFrame, SOMAMatrix, SOMACollection or SOMA. SOMACollection objects are typically persistent, and may be nested to any level (eg, SOMACollection may contain other SOMACollection objects). Keys in the map are unique and singular (no duplicates, ie, it is not a multi-map).
```
johnkerl, line 84: theoretically to any level but realistically when would we ever want to go beyond three? corpus -> multimodal -> each mode

johnkerl: does allowing (encouraging) arbitrary depth help us?

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+- a table schema, expressed as an `Arrow.RecordBatch` type, defining number of columns, and their respective names and types
+- the index column name
+
+A SOMADataFrame must contain at least a single column which is the named index column. The index column _must_ be of `string` type. Index values are non-null and must be unique (ie, no duplicates are allowed) within the object. The index is sparse and unordered, ie, supported indexing is by ID value only, with no indexing order or positional (offset-based) indexing.
+
+Most language-specific implementations will present SOMADataFrame in more convenient forms, such as Python `pandas.DataFrame`, R `data.frame` or other environment-specific data structures. Specific implementations of the SOMA API may extend this type with convertors to/from the underlying data structure, and enforce indexing conventions.
+
+> ⚠️ **Issue** - this spec requires that index column is a `string` type. Most underlying "engines" will support more flexibility. Do we want to enable that flexibility, or stick with requiring string for simplicity/commonality?
```
johnkerl, line 97: i prefer we just say string & hold to that ...

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+SOMAMatrix is a 2-D multi-column table of monomorphic arrays, ie, a "dataframe" with two index dimensions. SOMAMatrix is an extension of SOMADataFrame, with two indexing dimensions. The SOMAMatrix has a user-defined schema, which includes:
+
+- a table schema, expressed as an `Arrow.RecordBatch` type, defining the number of columns, and their respective names and types
+- the _primary_ index column name
+- the _secondary_ index column name
+
+A SOMAMatrix must contain (minimally) two columns which are the primary and secondary index columns. The index columns _must_ be of `string` type. Index values are non-null and must be unique (ie, no duplicates are allowed) within the object. The index is sparse and unordered, ie, supported indexing is by ID value only, with no indexing order or positional (offset-based) indexing.
+
+> ⚠️ **Issue** - SOMAMatrix and SOMADataFrame differ only in the number of dimensions (2 vs 1 respectively). Should they be unified into a single type object/API of more general nature? Clarity vs flexibility?
```
johnkerl, line 109: prefer 1 vs 2, keep them distinct

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+## SOMAMatrix
+
+SOMAMatrix is a 2-D multi-column table of monomorphic arrays, ie, a "dataframe" with two index dimensions. SOMAMatrix is an extension of SOMADataFrame, with two indexing dimensions. The SOMAMatrix has a user-defined schema, which includes:
+
+- a table schema, expressed as an `Arrow.RecordBatch` type, defining the number of columns, and their respective names and types
+- the _primary_ index column name
+- the _secondary_ index column name
+
+A SOMAMatrix must contain (minimally) two columns which are the primary and secondary index columns. The index columns _must_ be of `string` type. Index values are non-null and must be unique (ie, no duplicates are allowed) within the object. The index is sparse and unordered, ie, supported indexing is by ID value only, with no indexing order or positional (offset-based) indexing.
```
johnkerl, line 107: the question of how many attribute columns is omitted here -- maybe best --

i would go ahead & add that _nominally_ there will be a single attribute -- this will help the reader 'aha, i know what this is for'

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+> ⚠️ **Issue** - SOMAMatrix and SOMADataFrame differ only in the number of dimensions (2 vs 1 respectively). Should they be unified into a single type object/API of more general nature? Clarity vs flexibility?
+
+## SOMA
+
+The SOMA type is a specialized SOMACollection, representing an annotated 2-D matrix of observations across a set of variables. The SOMA composes SOMACollection, SOMADataFrame and SOMAMatrix types using a set of predefined fields (keys & values), to represent the annotated matrix. Each predefined field in the SOMA has a well-defined type, dimensionality and indexing convention. Other user defined data may be added to a SOMA, as a SOMA is a specialization of a SOMACollection.
```
johnkerl, line 113: here too -- keeping it loose/drafty -- just giving the reader an 'anchor' to hold onto while reading/absorbing/grokking -- mention we _might_ think of `SOMA` as "the thing" that has `obs`/`var`/`X` that single-cell savvy readers will see and head-nod at

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+Implementations _should_ enforce the naming and indexing constraints on these pre-defined fields. Pre-defined fields are distinguished from other user-defined collection elements, where no schema or indexing semantics are presumed or enforced.
+
+The predefined fields within SOMA share common indices and `shape`:
+
+- obs_id - the observation, a `string` type
+- var_id - the variable, a `string` type
+
+SOMA pre-defined fields are principally differentiated by:
+
+- dimensionality, eg, 1-D (SOMADataFrame) or 2-D (SOMAMatrix)
```
johnkerl, line 124: nice

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+- index column names, `obs_id` or `var_id` or both
```
johnkerl, line 125: if we're being specific about spelling `obs_id` and `var_id` then we may as well add `obs_id_i`/`obs_od_j` and `var_id_i`/`var_id_j` too

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+The pre-defined fields are:
+
+| Field name | Field type                              | Field description                                                                                                                                                                                                                                               |
+| ---------- | --------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `obs`      | `SOMADataFrame`                         | Primary annotations on the _observation_ axis. Must be of length #observations (ie, every observation _must_ be defined in this field), and must contain and be indexed by the `obs_id` column. The contents of `obs_id` define the _observation_ index domain. |
```
johnkerl, line 131: i really like how specific you are here. right here is where that's _most_ needed.

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+| `obsm`     | `SOMACollection[string, SOMADataFrame]` | Secondary _observation_ annotations, indexed with `obs_id`                                                                                                                                                                                                      |
+| `varm`     | `SOMACollection[string, SOMADataFrame]` | Secondary _variable_ annotations, indexed with `var_id`                                                                                                                                                                                                         |
+| `obsp`     | `SOMACollection[string, SOMAMatrix]`    | Pairwise _observation_ annotations, indexed with `[obs_1_id, obs_2_id]`                                                                                                                                                                                         |
```
johnkerl, line 135: ok here is fine -- above i said maybe we should say these explicitly -- you're doing it here -- this is fine

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+| `X`        | `SOMACollection[string, SOMAMatrix]`    | Sparse observations on a variable, indexed by `[obs_id, var_id]`                                                                                                                                                                                                |
+
+The `obs_id` domain for `obsp`, `obsm` and `X` (primary dimension) are the values defined by the `obs` SOMADataFrame `obs_id` column. The `var_id` domain for `varp`, `varm` and `X` (secondary dimension) are the values defined by the `var` SOMADataFrame `var_id` column. In other words, all predefined fields in the SOMA share a common `obs_id` and `var_id` domain, which is defined by the contents of the respective columns in `obs` and `var` SOMADataFrames.
+
+As with other SOMACollections, the SOMA also has a `metadata` field, and may contain other user-defined elements.
+
+### SOMA field constraints
+
+The following naming and indexing constraints are defined for the SOMA:
```
johnkerl, line 145: this section is just absolutely golden :grin:

brainstorming.md
johnkerl: ... i just read below & now see you use `SOMACollection` for things like containing the `obsm` element arrays so i retract this now ;)

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+The abstract data model comprises three "core" data types, and one specialization intended to improve ease of use and interoperability. The core types are:
+
+- SOMACollection - a string-keyed container (key-value map) of other SOMA data types, eg, SOMADataFrame, SOMAMatrix and SOMACollection.
+- SOMADataFrame - a 1-D, string-indexed, multi-column table, with monomorphic columns of equal length -- essentially a single-index dataframe.
+- SOMAMatrix - a 2-D, sparse, string-indexed multi-column matrix, with monomorphic columns of equal length -- essentially a two-index dataframe.
+
+In addition, a convenience type is defined:
+
+- SOMA - a specialization and extension of SOMACollection, codifying a set of naming and indexing conventions to represent an annotated, 2-D matrix of observations across a well-defined set of variables.
+
```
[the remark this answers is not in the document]
johnkerl: maybe add a quick note about how `SOMACollection` can be used to contain `obsm` elements etc

all throughout i've had the notion in my head that `SOMACollection` is _above_ SOMAs and it wasn't until several screenfuls down when i saw you using them (rightfully!) as containers for `obsm`/`varm` elements that i had the _aha_, a `SOMACollection` is a list/group _at any level_.

maybe help the reader have that same 'aha' sooner by an explicit foreshadow that `SOMACollection` can contains SOMAs, _and_ can contain things _within_ a SOMA

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+> ⚠️ **Issue** `raw` -- do we need `raw.X` and `raw.var` at all, or can we model them with existing structures? **Proposal**: the current `X` and `varm` can represent typical `raw` semantics (filtered subset, or transformed values). Assume that higher level usage conventions will utilize these to represent raw, eg, `X['raw']` and `varm['raw']`, or alternatively by adding additional columns to `var` that indicate filtering status.
+
+> ⚠️ **Issue** `obsm` and `varm` schema -- if these are RecordBatches indexed by obs_id / var_id, then they are essentially identical to obs/var slots. Are they necessary, or can we represent obsm/varm with conventions layered on top of obs/var? if they are different, how? The only current difference is that `obs` and `var` are required to have an entry for each value on that dimension, whereas `obsm`/`varm` do not require all dimension IDs to exist (ie, they are sparse like X). **Proposal**: retain, but differentiate by considering obsm/varm as secondary annotations on the axis, which do **not** require a value for each obs_id/var_id (ie are **sparse**).
```
johnkerl, line 161: i prefere them separately stored (to ease mental framing by all parties involved) but could be talked out of that with a compelling reason to overlap storage

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+> ⚠️ **Issue** `obsm` and `varm` names - field inspiration and name were borrowed from AnnData, but are misleading in this context as both are essentially a collection of SOMADataFrame values, not a collection of dense "matrix" in the AnnData sense. We should consider renaming them. While they can express the typical use case of AnnData.obsm/varm, they are better thought of as a collection of DataFrame containing secondary annotations for the obs/var axis. Possible ideas for better names: obsd/vard ('d' as in dataframe), obs_axis/var_axis, "axis annotation layers", or?
```
johnkerl, line 163: i defer 100% to the biopeople here

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+| obsp                    | Index column names are `['obs_1_id', 'obs_2_id']`. All `obs_1_id` and `obs_2_id` values exist in the `obs_id` column of the `obs` field.                                                 |
+| varp                    | Index column names are `['var_1_id', 'var_2_id']`. All `var_1_id` and `var_2_id` values exist in the `var_id` column of the `var` field.                                                 |
+| X                       | Index column names are `['obs_id', 'var_id']`. All `obs_id` values exist in the `obs_id` column of the `obs` field. All `var_id` values exist in the `var_id` column of the `var` field. |
+
+> ⚠️ **Issue** `raw` -- do we need `raw.X` and `raw.var` at all, or can we model them with existing structures? **Proposal**: the current `X` and `varm` can represent typical `raw` semantics (filtered subset, or transformed values). Assume that higher level usage conventions will utilize these to represent raw, eg, `X['raw']` and `varm['raw']`, or alternatively by adding additional columns to `var` that indicate filtering status.
```
johnkerl, line 159: i defer to the biopeople here

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+As any given storage "engine" upon which SOMA is implemented may have additional features and capabilities required to support advanced use cases, it is expected that SOMA implementations will expose engine-specific features. Where possible, these should be implemented to avoid conflict with future changes to the common SOMA API.
+
+> ⚠️ **Issue** - this spec needs to provide guidance on HOW to do that. Fro example, can we carve out some namespace for extensions in the functional API?
+
+In the following, a pseudo-object-oriented syntax is used, eg, `obj.op()` is calling `op()` upon `obj`. Where a type is used, eg, SOMA.op(), it is a `static` operation (ie, no state other than parameters provided).
```
johnkerl, line 173: good call -- various readers with strengths/weaknesses in various languages will be reading this & this is very welcome-mat-come-on-in :)

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+```
+  mapping.get(string key) -> value
+  mapping.has(string key) -> bool
+  mapping.set(string key, simple value) -> void
+  mapping.del(string key) -> void
+  mapping.__iterator__
+  mapping.__length__
+```
+
+> ⚠️ Issue - do we need a "is_readonly" or "is_immutable" attribute, or assume user can figure it out from the context?
```
johnkerl, line 194: 🤔

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+> ℹ️ **Note** - it is possible that the data model will grow to include more complex value types. If possible, retain that future option in any API defined.
+
+## SOMACollection
+
+In the following, `soco` refers to an SOMACollection instance or handle.
```
johnkerl, line 200: as above, `SOMACollection` has two roles -- 'outer' (contains SOMAs) and 'inner' (contains elements within a SOMA, e.g. `obsm`/`varm`)

i think these ops are necessary for both roles

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+```
+SOMACollection.create(uri) ->
+SOMACollection.open(uri, mode=read|write) -> soco
+SOMACollection.delete(uri) -> void
+SOMACollection.exists(uri) -> bool   # exists and is a SOMACollection
+soco.close()
+soco.uns -> metadata as a SOMAMapping
+```
+
+> ⚠️ **Issue** - do we need a specific open/close API, or is the object stateless? Or is that language/environment-specific?
```
johnkerl, line 211: i prefer that open/close be implementation details within the storage layer, so

* API level delete
* Storage level:
  * storage.open
  * storage.delete
  * storage.close
 
since SOMA is an API, not a storage spec, i think notions of open/close don't belong in the API -- open/close is a storage-level issue

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+Parameters:
+
+- ids - list of indices to read. Defaults to 'all'
+- column_names - the named columns to read and return. Defaults to all.
+- filter - [value filter](#value-filters) to apply (_TBD_). Defaults to no filter.
+
+> ⚠️ **Issue** - sdf.read() needs further definition for:
+>
+> - incremental/streaming queries - critical gap
+> - result ordering - would be useful, but also adds complexity (order by index? order by column value?). Most useful for streaming queries where the total result may be too large to load into memory and sort.
```
johnkerl, line 282: i need to read up on how `Arrow.RecordBatch` plays with things like `yield` paradigms & iteration over larger-than-RAM results ...

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+> - filtering based upon values - need to write the section: [Value Filter](#value-filters)
+
+> ℹ️ **Note** -- with this definition for read(), it should be possible to implement the equivalent of the higher level functions in the current prototype:
+>
+> - ids() -- read(ids=all, column_names=['the index column name, eg, obs_id'])
+> - query()/attribute_filter() -- read(column_names=['ids', 'a_col'], filter='a_col == "foobar"')
+
+### Method: write()
+
+Write an Arrow.RecordBatch to the persistent object. As duplicate index values are not allowed, index values already present in the object are overwritten and new index values are added.
```
johnkerl, line 292: good call to make

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+
+## SOMA
+
+SOMA supports all SOMACollection functions, including access to the predefined fields (eg, `X`) using SOMACollection API. The primary difference in behavior between a SOMACollection and a SOMA is that the SOMA _should_ enforce naming and indexing conventions for the predefine fields (eg, `obs`), as define in the [SOMA data model](#soma).
+
+Summary:
+
+```
+SOMA.create(uri) -> soma
+SOMA.open(uri, mode=read|write) -> soma
```
johnkerl, line 335: same as above -- i think open/close are storage-level concerns, not API-level concerns

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+soma.del(string key) -> void
+soma.__iterator__ -> key/value iterator for collection
+soma.__length__ -> int
+```
+
+## General Utilities
+
+Summary:
+
+```
```
johnkerl, line 355: this is the bomb :)

brainstorming.md
```diff
@@ -0,0 +1,372 @@
...
+get_storage_engine() -> string            # return underlying storage engine name, eg, "tiledb"
+```
+
+### Method: get_SOMA_version
+
+This is a pre-release specification in active development. As defined by [semver](https://semver.org/), this API is defined as version `0.0.0-dev`.
+
+## Value Filters
+
+> 🛑 **To be specified** -- a string-based language which can express simple filters and filter combinations based on column values, eg, "col_name == 'value'".
```
johnkerl, line 368: in R it's idiomatic to not use strings -- not
```
dplyr::filter(mtcars, cyl == 4)
```
but
```
dplyr::filter(mtcars, 'cyl == 4')
```

so it's high-level important that we not narrow down to "strings" here

let's keep it generic & say "query expressions" & that will be agnostic to the presence/absence of string quotes around whatever the query expression is

brainstorming.md
bkmartinjr: it does not address that. Arrow is just an in-memory typing and serialization system.  A RecordBatch data structure - a general purpose group of named/typed arrays of equal length.  Any decisions about how to use it are outside the scope of the data structure.

the value in using it is that we now can utilize the fact that many other packages (eg, Pandas, numpy, etc) can import/export arrow types without any further transcoding (no copy, no serialization, etc - very efficient).  Using it here gives us a language- and environment-neutral means of defining schema & result "structure" without overly binding to specific packages (eg, whatever is used in Python/R/Julia/Go/...)

brainstorming.md
johnkerl: or `SOMAAnnotationMatrixEnsemble`?

brainstorming.md
johnkerl: ... what i am now (having read through) thinking of as the 'outer role' and the 'inner role' which `SOMACollection plays both of

brainstorming.md
johnkerl: @bkmartinjr what if the `Arrow.RecordBatch` is too big (perhaps far too big) for memory? i think we will need _some_ way of indicating partial results ...

johnkerl: ... even if it's an indication that people will need to call `read` multiple times in some kind of 'continue-from-last' paradigm (if that's the way we go) ...

brainstorming.md
bkmartinjr: This idea of delegating (plus/minus some details) is interesting.  Couple of questions:
* what about `exists()` and related?
* what about `create()`?
* I'm not sure I understand `api.delete` above.  Do you mean `update` of existing objects (as in SOMACollection  add/del and SOMADataFrame read/write)?

I think you are suggesting that the SOMACollection, SOMADataFrame and SOMAMatrix lifecycle (create, delete, open, close) are entirely outside the scope of the SOMA API.

brainstorming.md
bkmartinjr: > what if the Arrow.RecordBatch is too big

See the previous bullet:  `- incremental/streaming queries`, which is the same as `partial results`

the RecordBatch is just a data structure holding the result of the query.   If the query/read is too big, it will fail.  So you are right that we need a way to `page` through the results.  I have not yet thought through how to best present that, which is why I listed it as an open (and critical) gap

brainstorming.md
aaronwolen: Agreed. We just had a [similar conversation](https://github.com/TileDB-Inc/tiledbsc/pull/67#discussion_r906708148) about this very topic in the context of attempting to keep the r/py slicing APIs consistent but still idiomatic to the underlying language.

brainstorming.md
johnkerl: @bkmartinjr awesome! then we're both aware that we need to articulate (even at high level) something about how we want paging to feel like

brainstorming.md
bkmartinjr: `simple` is defined in the section titled `Base Type System`.  I will add a link in the spec to clarify.

brainstorming.md
bkmartinjr: @aaronwolen - this is a key-value store, ie, schema-free metadata. The only difference (currently) from `uns` is the requirement that all values in the store be primitives or strings (and it wouldn't surprise me if this evolves over time).

I personally prefer `metadata` and will integrate that change in the next rev.  However, I'm also very happy to change to something better if we find one....

brainstorming.md
bkmartinjr: Good catch - we need to add clarity here.

BTW, in this API, there is no concept of an "attribute column" (I assume you are using that in the TileDB sense of the term).  Rather, there are N>=3 columns, and two of them are designated the index columns.  Underlying implementations need to figure out how to map that to their schema (which in the TileDB engine would simply mean the two named index columns are dims, the rest are attributes).

brainstorming.md
bkmartinjr: @johnkerl - could you clarify your comment?  I'm not understanding what the concern is...  Thanks!

brainstorming.md
bkmartinjr: this is specified in the constraints section.

brainstorming.md
bkmartinjr: agree, and that was my intent by using the same type name in both cases, with the same semantics, operations, etc.

Did you see a place where this was incorrectly documented, or is a request to add more explicit prose?

brainstorming.md
bkmartinjr: In the next revision, I will attempt to thread the needle by specifying the operators and capabilities of the push-down filters, but will delegate the syntax to per-language specs.

brainstorming.md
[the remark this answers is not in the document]
bkmartinjr: I'll add some prose to help with this.

brainstorming.md
bkmartinjr: you have the essence of this right (or at least as I intended it).  Both use cases are persistent, and the flexibility to nest will be important for some use cases (eg, "atlas" of "multi-modal", etc).

I'm adding a bit of expository prose as you suggested to make this clearer.

brainstorming.md
bkmartinjr: I will add some prose as you suggest.  I do not think your literal suggestion is what we want, as a new reader will have no idea what `soma.X` is or implies.  I'll state something in more standard English -- see if it is sufficient for a new reader...

brainstorming.md
bkmartinjr: @ambrosejcarr - would love your thoughts on this.

brainstorming.md
bkmartinjr: @johnkerl - please resolve this thread if you are OK with the changes to address this concern.

brainstorming.md
bkmartinjr: Fix added to both SOMADataFrame and SOMAMatrix.

brainstorming.md
bkmartinjr: Prose added to the SOMACollection section.

brainstorming.md
bkmartinjr: In thinking about this some more, I do not think the abstract SOMA spec should specify the _precise_ API for delayed and incremental reads.  Rather, it should say that it is available, and delegate to language-specific SOMA specs, as each environment will want to do it slightly differently.  I will add some proposed text to this effect.

johnkerl: it needn't be _precise_ but i think we should specify _something_ or different impls will do wildly different things.

* does a larger-than-memory read return some kind of iterator which yields batches?
* does it return a single batch with a 'there is more' flag (like many restful web apis)?

brainstorming.md
bkmartinjr: Apologies!  I misunderstood the original comment. We are aligned on intent - I will improve the prose.

brainstorming.md
bkmartinjr: obsolete issue given latest discussions.

brainstorming.md
bkmartinjr: removed from latest

brainstorming.md
bkmartinjr: I am resolving this thread as the latest doc contains:
* use of Arrow for base type system and read/write ABI
* a more explicit discussion that clarifies the use of RecordBatch/Tensor as the ABI for a read/write "chunk", and defers the actual codification of a delayed IO iterator to the per-language bindings (ie, this doc just proposes that it is possible to do incremental reads, but doesn't specify how the R or Python bindings will express that).

brainstorming.md
```diff
@@ -0,0 +1,508 @@
+# Introduction
+
+> **Status**: brainstorming, pre-proposal
+>
+> ℹ️ **Note** - this is an early draft and has not had any substantive review or circulation. It may contain controversial, undecided, or just plain wrong-headed ideas. I expect several major iterations will be required to converge the underlying primitives with use cases. Please see `issues` noted throughout the doc for a list of active discussions/debates. In particular, this doc has many known gaps identified throughout the with callouts.
+
+The goal of SOMA (“stack of matrices, annotated”) is a flexible, extensible, and open-source API providing access to annotated, 2D matrix data stored in multiple underlying formats and systems. The vision for this API family includes:
+
+- support access to persistent, cloud-resident datasets,
+- enable use within popular data science environments (eg, R, Python), using the tools of that environment (eg, Python Pandas integration),
```
johnkerl, line 10: i'll leave word-smithing large & small (example of the latter: _eg_ -> _e.g._) until separate review

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+
+> ⚠️ **Issue** - are there parts of the Arrow type system that we wish to _explicitly exclude_ from SOMA? I have left this issue open (ie, no specific text) for now, thinking that we can come back and subset as we understand what complex types are required, and how much flexibility should be in this spec. We clearly need some complex types (eg, RecordBatch, List, etc) as they are implied by `string`, etc. My own preference would be to mandate a small set of primitive types, and leave the rest open to implementations to support as they feel useful.
+
+> ⚠️ **Issue** - the above uses Arrow `string`. Should we be using `large_string` instead (no 2GB cap)?
+
+## Metadata
+
+All SOMA objects may be annotated with a small amount of metadata:
+
+- `metadata` (map[string, simple]) - user metadata, specified as string-keyed mapping to any [**simple**](#type-definitions-used-in-this-document) type (eg, int32, string, etc).
```
johnkerl, line 70: i know anndata often has metadata values which are lists of strings -- so defining "simple" to exclude "list of string" is, i believe, going to block the the ingest-anndata use-case immediately -- this known, concrete use-case warrants the complexity of supporting list of string in metadata

as a workaround, maybe map list of strings `["a","b","c"]` from other storages into comma-joined `"a,b,c"` -- ? but i prefer including support for list-of-string in our bare minimum

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+## Foundational Types
+
+The foundational types represent the core data structures used to store and index data. They are intended to be moderately general purpose, and to serve as building blocks for the [Composed Types](#composed-types) which codify domain-specific use cases (eg, single cell experimental datasets).
+
+### SOMACollection
+
+SOMACollection is an unordered, `string`-keyed map of values. Values may be any SOMA foundational or composed type, including other (nested) SOMACollection objects. Keys in the map are unique and singular (no duplicates, ie, the SOMACollection is _not_ a multi-map). The SOMACollection is expected to be used for a variety of use cases:
+
+- as a container of independent objects (eg, a collection of single-cell datasets, each manifest as a [SOMAExperiment](#soma-experiment) object)
+- as the basis for building other composed types (eg, using SOMACollection to organize pre-defined fields in [SOMAExperiment](#soma-experiment) such as multiple layers of `X`).
```
johnkerl, line 87: ❤️

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+
+## Foundational Types
+
+The foundational types represent the core data structures used to store and index data. They are intended to be moderately general purpose, and to serve as building blocks for the [Composed Types](#composed-types) which codify domain-specific use cases (eg, single cell experimental datasets).
+
+### SOMACollection
+
+SOMACollection is an unordered, `string`-keyed map of values. Values may be any SOMA foundational or composed type, including other (nested) SOMACollection objects. Keys in the map are unique and singular (no duplicates, ie, the SOMACollection is _not_ a multi-map). The SOMACollection is expected to be used for a variety of use cases:
+
+- as a container of independent objects (eg, a collection of single-cell datasets, each manifest as a [SOMAExperiment](#soma-experiment) object)
```
johnkerl, line 86: ❤️

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+- as the basis for building other composed types (eg, using SOMACollection to organize pre-defined fields in [SOMAExperiment](#soma-experiment) such as multiple layers of `X`).
+
+### SOMADataFrame
+
+`SOMADataFrame` is a multi-index, multi-column table of monomorphic arrays, all of equal length. A `SOMADataFrame` has a user-defined schema, which defines the number of columns and their respective column name and value type. The schema is expressed as an Arrow `Schema`. All `SOMADataFrame` contain a "pseudo-column" called `__rowid`, of type `uint64` and domain `[0, #rows)`. The `__rowid` pseudo-column contains a unique value for each row in the `SOMADataFrame`, and is intended to act as a join key for other objects, such as a `SOMANdArray`.
```
johnkerl, line 91: did we define _monomorphic_ anywhere? since this is a doc about an API/standard, we should spell out what we mean to remove all doubt

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+
+A `SOMADataFrame` may be `user-indexed` or `row-indexed`:
```
johnkerl, line 93: ✅

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+
+- `row-indexed` - each row is addressable (indexed) by the pseudo-column `__rowid` (ie, a single-index using offset indexing).
+- `user-indexed` - each row is addressable (indexed) by one or more user-specified index columns, which must be named columns in the dataframe schema.
+
+Row-indexed datafames must have no user-specified index columns and at least one column defined in their schema. User-indexed dataframes must have at least one schema-defined column named as an index column. The user-specified schema may not redefine the `__rowid` pseudo-column.
+
+Most language-specific bindings will provide convertors between SOMADataFrame and other convenient data structures, such as Python `pandas.DataFrame`, R `data.frame`.
+
+> ℹ️ **Note** - on a TileDB implementation, `row-indexed` SOMADataFrames will likely map to a _dense_ array, with a single uint64 dimension called **rowid, with domain [0 #nrows). A `user-indexed` SOMADataframe will be a _sparse_ array, with a dimension for each user-specified index column, and a hidden **rowid attribute.
```
johnkerl, line 102: ✅

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+
+> ℹ️ **Note** - on TileDB this is an sparse array with `N` uint64 dimensions of domain [0, maxUint64), and a single attribute.
+
+## Composed Types
+
+Composed types are defined as a composition of foundational types, adding name, type and indexing constraints. These types are intended to facilitate data interoperability, ease of use, and _potentially_ enable implementation optimizations by virtue of their typing and structural guarantees. The initial composed types are motivated by single cell biology, but additional types may be added in the future for more diverse use cases.
+
+### SOMAExperiment & SOMAMeasurements
+
+`SOMAExperiment` is a specialized `SOMACollection`, representing an annotated 2-D matrix of measurements. In the single-cell biology use case, a `SOMAExperiment` can represent multiple modes of measurement across a single collection of cells. Within a `SOMAExperiment`, a set of measurements on a single set of variables (features) is represented as a `SOMAMeasurements`.
```
johnkerl, line 132: perhaps nitty: `SOMAMeasurements` already being plural is awkward. in the presence of multiple assays, there are ... SOMAMeasurementses? i'd love for us to have a word which is singular for one dict-slot of `SOMAExperiment`. Whether `SOMAMeasurement`, `SOMAAssay`, `SOMAMeasuremntSet`, `SOMAMeasurementItem`, or something else -- something singular.

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+| `varp` collection values                           | All matrixes must have the shape `(#var, #var)`. The domain of both dimensions is the values of `var.__rowid`.                                                                                                                       |
+| `var_ms` collection values                         | All dataframes must have the shape `(#var,)`. The domain of the dimension is the values of `var.__rowid`.                                                                                                                            |
+
+> ⚠️ **Issue** `raw` -- do we need `raw.X` and `raw.var` at all, or can we model them with existing structures? **Proposal**: the current `X` and `varm` can represent typical `raw` semantics (filtered subset, or transformed values). Assume that higher level usage conventions will utilize these to represent raw, eg, `X['raw']` and `varm['raw']`, or alternatively by adding additional columns to `var` that indicate filtering status (eg, `var['is_filtered']`).
```
johnkerl, line 197: i think (disclaimer: "i am not a biologist") we need to decide where to put raw sooner than later. it's a central, immediate use-case and we just need to decide and stick.

as this PR has evolved, it's been marvelous to see our thinking evolve and crystallize into a well-thought-out way to handle multimodal data -- a pinch-point for cross-language/cross-tool operability -- to the point that we have names and a diagram which demonstrates our well-thought-out plan.

we need to achieve the same for raw.

your suggestion here is good -- I've no action item _for you_ on this -- my input is that we, collectively, need accept/counterpropose on this great suggestion of yours, sooner than later.

one concern i have (again "i am not a biologist") is what if a single experiment, with more than one assay type, has two different assays each of which has raw data? i'm not enough of a biologist to answer that question but i'm enough of an engineer to ask it ...

brainstorming.md
```diff
@@ -0,0 +1,508 @@
...
+
+### Method: get_SOMA_version
+
+This is a pre-release specification in active development. As defined by [semver](https://semver.org/), this API is defined as version `0.0.0-dev`.
+
+## Value Filters
+
+> ⚠️ **To be further specified**
+
+Value filters are expressions used to filter the results of a `read` operation, and specify which results should be returned. The specific means to create and manipulate a value filter is delegated to per-language specifications. This specification uses a pseudo-language for _examples only_.
```
johnkerl, line 474: this is a good balance of specific, and open-ended

brainstorming.md
johnkerl, line 132: this is very nice

a little thing: you're saying "multiple modes of measurement" which is good

please include the word "multimodal" explicitly. different readers are going to come into this doc for various reasons, and will read at different depths from skim to search to grok. anyone who comes in asking "hey, what does this new 'SOMA' spec say about multimodal?" needs to be able to find that answer and see that, yes, we are handling it authoritatively and confidently.

johnkerl (commented):
Initial thoughts on first read of today's rev --

* I left out copy-editing per se, except a couple places where I thought some wording was particularly important
* On your request, I limited the amount of positive, non-action-inducing feedback -- rest assured I'm smiling and fist-pumping multiple times as I read this, and please know my feedback is almost entirely positive
* Of course, I defer much to our biofriends who will know more
* TL;DR I love the concrete proposal for multimodal & I think it will work. Number-one thing needing ideation is how to get raw-handling up to that level.

brainstorming.md
bkmartinjr: done.

brainstorming.md
bkmartinjr: keep it up and you will be assigned editorial responsibilities :-(

johnkerl: @bkmartinjr tbh I enjoy it! :) Used to work as a copy-editor for a brief while, & was happy at it :)

brainstorming.md
johnkerl: > Besides "short list of simple types", what else do you recommend as a "minimum" requirement?

@bkmartinjr my minimum list includes your simples, and list thereof

brainstorming.md
bkmartinjr: will add!

brainstorming.md
bkmartinjr: I defer to @ambrosejcarr who owns naming!

The original proposal was `SOMAMesurementSet`, which satisfies your criteria.  I am happy to adopt that.  The change to `SOMAMesurements` was solely to recognize that the type is not a true set in the mathematical or data structure sense of the word.

@johnkerl - do you think `SOMAMeasurement` or `SOMAMeasurementSet` is best (those are the two I'm comfortable with)?  I do not think `SOMAAssay` is good - it overloads the meaning with unnecessary specificity IMHO.

brainstorming.md
bkmartinjr: you have a calling!

For now, lets focus on those issues that are misleading, block comprehension, etc.  There will definitely be a cleanup and "add examples" pass at some point, after we agree on the proposal.  I will be _very_ happy to have assistance at that point, as I am (demonstrably) an amateur at this "writing thing".

bkmartinjr: "eg" and "ie" are fixed.

brainstorming.md
bkmartinjr: Not all lists can be expressed as an Arrow array or table. For example, I believe that they need to be monomorphic.

My suggestion:  anything that can be stored as an Arrow Array is "OK".  Indicate that they should be "small", and leave the definition of "small" up to the implementation.

Thoughts?

johnkerl: i am happy with us supporting arrays of ints and arrays of strings in metadata as i know that already in my data-explorations i've seen lots of both in the wild -- we should support these from day one

brainstorming.md
bkmartinjr: sold - will make this change.

brainstorming.md
bkmartinjr: Added to the issues section:
> 5. Should `raw` vs `processed` conventions be codified in this spec (eg, as predefined fields in SOMAExperiment/SOMAMeasurement)? Or should they be the purview of higher level schema/conventions, built upon the general-purpose features of this API?

brainstorming.md
bkmartinjr: I have added Arrow arrays to the list of types supported by metadata (values).  Keys are still strictly strings.

johnkerl (approved):

brainstorming.md
```diff
@@ -169,7 +167,7 @@ The `SOMAMeasurement` is a sub-element of a SOMAExperiment, and is otherwise a s
 
 For the entire `SOMAExperiment`, the index domain for the elements within `obsp`, `obsm`, `obs_ms` and `X` (first dimension) are the values defined by the `obs` `SOMADataFrame` `__rowid` column. For each `SOMAMeasurement`, the index domain for `varp`, `varm`, `var_ms` and `X` (second dimension) are the values defined by the `var` `SOMADataFrame` `__rowid` column in the same measurement set. In other words, all predefined fields in the `SOMAMeasurement` share a common `obsid` and `varid` domain, which is defined by the contents of the respective columns in `obs` and `var` SOMADataFrames.
 
-As with other SOMACollections, the `SOMAExperiment` and `SOMAMeasurement` also have a `metadata` field, and may contain other user-defined elements.
+As with other SOMACollections, the `SOMAExperiment` and `SOMAMeasurement` also have a `metadata` field, and may contain other user-defined elements. Keys in a `SOMAExperiment` and `SOMAMeasurement` beginning with the characters `_`, `.`, or `$` are reserved for ad hoc use, and will not be utilized by this specification. All other keys are reserved for future future specifications.
```
johnkerl, line 170: future future -> future

bkmartinjr: thank you!  fixed.

brainstorming.md
ambrosejcarr: I can think of lots of examples where complex metadata _exist_ (lists or arrays of strings, json blobs) but none where complex metadata are _required_. 

> i know anndata often has metadata values which are lists of strings -- so defining "simple" to exclude "list of string" is, i believe, going to block the the ingest-anndata use-case immediately

I think this is the right problem to focus on. Because AnnData is more permissive, we're often going to either (a) fail ingestion or (b) introduce data loss. I favor loss of data + a warning during ingestion when that occurs is a good approach given the cases I'm aware of.

brainstorming.md
ambrosejcarr: I didn't have strong feelings between `SOMAMeasurement` and `SOMAMeasurementSet`. I find the argument against `Set` compelling, so am happy with the former.

brainstorming.md
bkmartinjr: In offline discussion of this, it was also pointed out that the SOMACollection is the basis of all composed types, and it can store arbitrary dataframe/array data with a user-specified name. This is a relatively full-featured substitute for complex metadata, and provides the affordances of those more advanced types (eg, incremental read/write).  Decision taken to revert the metadata back to simple/primitive types.

ivirshup: > This is a relatively full-featured substitute for complex metadata,

This is basically where AnnData has landed for now too.

However we're going to expand the per element metadata within anndata. This is needed for a number of use cases around spatial data, and could be quite useful for handling provenance. Right now, I'm thinking this metadata must be json serializable.

bkmartinjr: > we're going to expand the per element metadata within anndata

Can you elaborate on what way you will expand it?  Do you mean adding a `uns` equivalent to more objects (for example, `adata.obsm.uns['foo'] = 'bar'`?  Or that the supported type system for `uns` values will expand?  Or something else?

Really appreciate your engagement!

bkmartinjr: Many thanks for the concrete example.

This example (JSON serialized string on an obsm array) fits our current design thinking (string/primitive metadata values on all array/dataframe objects).

brainstorming.md
```diff
@@ -0,0 +1,557 @@
...
+| delete(uri)                             | Delete the SOMADataFrame specified with the URI.                                        |
+| exists(uri) -> bool                     | Return true if object exists and is a SOMADataFrame.                                    |
+| get metadata                            | Access the metadata as a mutable [`SOMAMetadataMapping`](#SOMAMetadataMapping)          |
+| get type                                | Returns the constant "SOMADataFrame"                                                    |
+| get shape -> (int, ...)                 | Return length of each dimension, always a list of length `ndims`                        |
+| get ndims -> int                        | Return number of index columns                                                          |
+| get schema -> Arrow.Schema              | Return data schema, in the form of an Arrow Schema                                      |
+| get is user-index -> bool               | Return true if user-indexed, false if row-indexed.                                      |
```
johnkerl, line 237: `get_is_user_indexed`

bkmartinjr: I cleaned these up so that they are consistent in the doc.  TY!

brainstorming.md
```diff
@@ -0,0 +1,593 @@
...
+### SOMACollection
+
+SOMACollection is an unordered, `string`-keyed map of values. Values may be any SOMA foundational or composed type, including other (nested) SOMACollection objects. Keys in the map are unique and singular (no duplicates, i.e., the SOMACollection is _not_ a multi-map). The SOMACollection is expected to be used for a variety of use cases:
+
+- as a container of independent objects (e.g., a collection of single-cell datasets, each manifest as a [SOMAExperiment](#soma-experiment) object)
+- as the basis for building other composed types (e.g., using SOMACollection to organize pre-defined fields in [SOMAExperiment](#soma-experiment) such as multiple layers of `X`).
+
+### SOMADataFrame
+
+`SOMADataFrame` is a multi-column table with a user-defined schema, defining the number of columns and their respective column name and value type. The schema is expressed as an Arrow `Schema`. All `SOMADataFrame` contain a "pseudo-column" called `__rowid`, of type `uint64` and domain `[0, #rows)`. The `__rowid` pseudo-column contains a unique value for each row in the `SOMADataFrame`, and is intended to act as a join key for other objects, such as a `SOMADenseNdArray`.
```
johnkerl, line 86: i don't see any requirement the row IDs be consecutive starting at 0.

if we require that we should explicitly say it. if we don't require it we should point out that fact, as clarification to any readers who might have assumed otherwise.

note that since `obs` and `X` are joined by row IDs, _if_ `X` is dense then it's definitely 0-up and contiguous, which _forces_ `obs` to be 0-up and contiguous. however, if `X` is sparse, i don't see any language in this spec that would tell a reader whether `obs` must also be 0-up and contiguous, or not.

brainstorming.md
```diff
@@ -0,0 +1,593 @@
...
+Most language-specific bindings will provide convertors between SOMADataFrame and other convenient data structures, such as Python `pandas.DataFrame`, R `data.frame`.
+
+### SOMADenseNdArray
+
+`SOMADenseNdArray` is a dense, N-dimensional array with offset (zero-based) integer indexing on each dimension. The `SOMADenseNdArray` has a user-defined schema, which includes:
+
+- type - a `primitive` type, expressed as an Arrow type (e.g., `int64`, `float32`, etc), indicating the type of data contained within the array
+- shape - the shape of the array, i.e., number and length of each dimension
+
+All dimensions must have a positive, non-zero length.
```
johnkerl, line 97: for completeness (as long as we're saying this) then we should also say that we require that the number of dimensions be positive (N > 0).

brainstorming.md
bkmartinjr: Actually, it is specified in this paragraph but perhaps too tersely.   This text states exactly what you are asking for:

> `__rowid`, of type `uint64` and domain `[0, #rows)`

I will add additional clarification

brainstorming.md
bkmartinjr: resolved

brainstorming.md
```diff
@@ -0,0 +1,593 @@
...
+
+### SOMADataFrame
+
+`SOMADataFrame` is a multi-column table with a user-defined schema, defining the number of columns and their respective column name and value type. The schema is expressed as an Arrow `Schema`. All `SOMADataFrame` contain a "pseudo-column" called `__rowid`, of type `uint64` and domain `[0, #rows)`. The `__rowid` pseudo-column contains a unique value for each row in the `SOMADataFrame`, and is intended to act as a join key for other objects, such as a `SOMADenseNdArray`.
+
+Most language-specific bindings will provide convertors between SOMADataFrame and other convenient data structures, such as Python `pandas.DataFrame`, R `data.frame`.
+
+### SOMADenseNdArray
+
+`SOMADenseNdArray` is a dense, N-dimensional array with offset (zero-based) integer indexing on each dimension. The `SOMADenseNdArray` has a user-defined schema, which includes:
```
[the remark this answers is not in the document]
bkmartinjr: done!

brainstorming.md
```diff
@@ -82,24 +83,15 @@ SOMACollection is an unordered, `string`-keyed map of values. Values may be any
...
-
-A `SOMADataFrame` may be `user-indexed` or `row-indexed`:
-
-- `row-indexed` - each row is addressable (indexed) by the pseudo-column `__rowid` (i.e., a single-index using offset indexing).
-- `user-indexed` - each row is addressable (indexed) by one or more user-specified index columns, which must be named columns in the dataframe schema.
-
-Row-indexed datafames must have no user-specified index columns and at least one column defined in their schema. User-indexed dataframes must have at least one schema-defined column named as an index column. The user-specified schema may not redefine the `__rowid` pseudo-column.
+`SOMADataFrame` is a multi-column table with a user-defined schema, defining the number of columns and their respective column name and value type. The schema is expressed as an Arrow `Schema`. All `SOMADataFrame` contain a "pseudo-column" called `__rowid`, of type `uint64` and domain `[0, #rows)`. The `__rowid` pseudo-column contains a unique value for each row in the `SOMADataFrame`, and is intended to act as a join key for other objects, such as a `SOMADenseNdArray`.
```
stavrospapadopoulos: Just noting here that there are two columns of note:
1. The `__rowid` mentioned here, which is *materialized* and is the "join" column with some *array* dimension values.
2. A virtual "row id" column for the dataframe indicating the positions of the rows in the dataframe, which is *not materialized* and simply facilitates slicing dataframe rows by position.

Not sure if you'd like to mention both here and give different descriptive names to avoid confusion.

stavrospapadopoulos (approved):
Just one comment, otherwise looks good to me. I think this is good to start with, the first implementation may expose potential edits to the spec. Thanks @bkmartinjr!

brainstorming.md
bkmartinjr: resolved in the latest spec per our conversation.  Please have a read and see what you think.

brainstorming.md
```diff
@@ -0,0 +1,689 @@
...
+Create a new SOMADataFrame with user-specified URI and schema.
+
+```
+create(string uri, Arrow.Schema schema, platform_config) -> void
+```
+
+Parameters:
+
+- uri - location at which to create the object
+- schema - an Arrow Schema defining the per-column schema. This schema must define all columns, including columns to be named as index columns. All column names beginning with `soma_` are reserved. If the schema includes types unsupported by the SOMA implementation, an error will be raised.
```
johnkerl, line 267: > , including columns to be named as index columns

please remove

bkmartinjr: done

brainstorming.md
```diff
@@ -0,0 +1,691 @@
...
+### SOMABatchFormat
+
+Array read operations can return results in a variety of formats. The `SOMABatchFormat` format indicates the format encoding.
+
+| Batch format   | Description                                                                                           |
+| -------------- | ----------------------------------------------------------------------------------------------------- |
+| `dense`        | Return the coordinates of the slice (e.g. origin, shape) and an Arrow Tensor containing slice values. |
+| `coo`          | Return an Arrow.SparseCOOTensor                                                                       |
+| `csr`          | Return an Arrow.SparseCSRTensor                                                                       |
+| `csc`          | Return an Arrow.SparseCSCTensor                                                                       |
```
ivirshup, line 609: Just wanted to drop in on this:

As I've mentioned before (https://github.com/single-cell-data/SOMA/issues/18#issuecomment-1075398109) I'm a little concerned about arrow's level of support for its Tensor types, especially the sparse tensor types. Last I looked into this, it seemed like there were no tests in the arrow ipc suite for these. E.g. there's no confirmation interchange works for tensors. I believe this has led to a number of implementations just skipping out on tensor (esp. sparse tensor) support.

That said, I would really love it if these were better supported.

bkmartinjr: I hear ya...that said, I am unsure there is any good (portable, common, etc) alternative other than removing the capability (ie, only supporting COO in a RecordBatch) _or_ defining our own data structure. IMHO, both are worse than taking the risk of being an early SparseTensor adoptor.

If I'm missing an alternative, would love to hear your ideas!

ivirshup: The only alternative I can think of is how vitessce handles sparse zarr AnnData matrices.

In effect, I'm not sure how different this is than defining a custom data structure, since clients will have to handle decoding.

My ideal here would be that there is some contact with the arrow team, there's some work on what they think about it, and some ownership of these structures goes to people who care about them. Tests could get added to the IPC, so implementation start supporting these (for example).

Following from SciPy this year, some discussions have started around binary formats for sparse data. Discussions are being recorded here: https://github.com/GraphBLAS/binsparse-specification and is largely w/ scientific-python and graphBLAS. Still quite early though.

brainstorming.md
```diff
@@ -1,6 +1,6 @@
 # Introduction
 
-> **Status**: brainstorming, pre-proposal
+> **Status**: brainstorming, pre-proposal, in active development. Version: `0.1.0-dev`
```
johnkerl, line 3: My preference: we call this new spec`0.2.0-dev` and we call the old spec `0.1.0-dev` since we already have releases named `0.1.x` for the old spec at
https://github.com/single-cell-data/TileDB-SOMA/releases/tag/0.1.11
https://github.com/TileDB-Inc/tiledbsc/releases

bkmartinjr: I"m happy to do this, but we can not make the mistake of conflating the API / spec version with the implementation version.  They will evolve independently.  So this is a one-time change to burn another version #.

bkmartinjr: Also, the spec had two different names for this function - fixed to use `get_SOMA_version`

johnkerl (approved):
🚢

PR #36, opened 2022-09-29
Add the very basic skeleton where the Python spec will go.

thetorpedodog:
This sets up the infrastructure for where we can put the base classes and whatever other Python stuff we need as we develop the SOMA spec. It's about the simplest possible Python project with the minimal amount of needed stuff.

---

I’m not too attached to the name, just that `somabase` was my best idea since `soma` and `singlecell` are both already taken.

### DIFF
--- a/python-spec/.gitignore
+++ b/python-spec/.gitignore
@@ -1,0 +1,2 @@
+**/__pycache__
+**/*.egg-info
--- a/python-spec/pyproject.toml
+++ b/python-spec/pyproject.toml
@@ -1,0 +1,8 @@
+[build-system]
+requires = ["setuptools"]
+build_backend = "setuptools.build_meta"
+
+[project]
+name = "somabase"
+description = "Python-language API specification and base utilities for implementation of the SOMA system."
+version = "0.0.0"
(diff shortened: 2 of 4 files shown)


bkmartinjr (approved):
Looks good to me.  Might want to also run by @johnkerl, @gspowley and @aaronwolen in case they have comments on the approach.

aaronwolen (approved):
LGTM. Thanks @thetorpedodog!

johnkerl:
Sorry @thetorpedodog I just now saw this (post-merge).

I'm super-confused.

Are we duplicating https://github.com/single-cell-data/TileDB-SOMA/tree/main/apis/python ?

johnkerl:
@thetorpedodog sorry, I saw this after the merge. I would have requested changes had I gotten here in time.

This is a great idea! TBH I didn't understand it, from this PR. Based on some side conversations I think this is clearer to me now (and, again, this is great)! Let's please do the following:

* Update the Description field of this PR
* Clarify that this is a spec-only class
* We plan to use `ABC` as mixins
* `tiledbsoma` (https://github.com/single-cell-data/TileDB-SOMA/tree/main/apis/python) will `import somabase`

PR #39, opened 2022-10-04
Add concrete classes and enums for options.

thetorpedodog:
Adds the "common interfaces" `IOfN`, `BatchSize`, and `BatchFormat`. These classes are the exact values that implementations will use, i.e. an implementation of the SOMA API will not implement its own `IOfN` or any of these other “option” types.

These types are implemented using the `attrs` package, essentially an enhanced version of the built-in `dataclasses` package. It is very lightweight and already extensively used within the Python ecosystem so it will not add much (if anything) to the size of a typical installed environment.

---

I started on the API structure for the various NDArray types and found that we needed some options before then.

The major decision here was the way I chose to implement `BatchSize`. Instead of e.g. two different types (like maybe `CountBatch(n)` and `BytesBatch(size)`) I think a single class is a bit more convenient for authors. If for some reason in the future we have a different type of batch size that has its own complex setup, the type itself can always be renamed or otherwise refactored and `BatchSize` can remain a factory function for compatibility.

An **open question**: For the `DENSE` type, what are the details of the “coordinates of the slice (e.g. origin, shape) and an Arrow Tensor containing slice values”?

(I would be inclined to specify that it should return data as a `namedtuple` (not an attrs class, so that users who want to just unpack it like `coords, tensor = some_call_which_returns_dense(...)`. But what would those elements be?)

### DIFF
--- a/python-spec/src/somabase/options.py
+++ b/python-spec/src/somabase/options.py
@@ -1,0 +1,95 @@
+"""Enums and other types used as options across methods of many types.
+
+These types are *concrete* and should be accepted as inputs
+"""
+
+import enum
+from typing import Optional
+
+import attrs
+
+
+@attrs.define(frozen=True)
+class IOfN:
+    """Specifies that a read should return partition ``i`` out of ``n`` total.
+
+    For a read operation that returns ``n`` partitions, the read operation will
+    return the ``i``th partition (zero-indexed) out of ``n`` partitions of
+    approximately equal size.
+    """
+    __slots__ = ()
+
+    i: int = attrs.field()
+    """Which partition to return (zero-indexed)."""
+    n: int = attrs.field()
+    """How many partitions there will be."""
+
+    @i.validator
+    def _validate(self, _, __):
+        del _, __  # Unused.
+        if not 0 <= self.i < self.n:
+            raise ValueError(
+                f"Partition index {self.i} must be in the range [0, {self.n})"
+            )
+
+
+@attrs.define(frozen=True)
+class BatchSize:
+    """Specifies the size of a batch that should be returned from reads.
+
+    Read operations on foundational types return an iterator over "batches" of
+    data, enabling processing of larger-than-core datasets. This class allows
+    you to control what the size of those batches is.
+
+    If none of these options are set, a "reasonable" batch size is determined
+    automatically.
+
+    For example::
+
+        BatchSize(count=100)
+        # Will return batches of 100 elements.
+
+        BatchSize(bytes=1024 ** 2)
+        # Will return batches of up to 1 MB.
+
+        BatchSize()
+        # Will return automatically-sized batches.
+    """
+    __slots__ = ()
+
+    count: Optional[int] = attrs.field(default=None)
+    """``arrow.Table``s with this number of rows will be returned."""
+    bytes: Optional[int] = attrs.field(default=None)
+    """Data of up to this size in bytes will be returned."""
+
+    @count.validator
+    @bytes.validator
+    def _validate(self, attr: attrs.Attribute, value):
+        if not value:
+            return  # None (or 0, which we treat equivalently) is always valid.
+        if value < 0:
+            raise ValueError(f"If set, '{attr.name}' must be positive")
+        if self.count and self.bytes:
+            raise ValueError("Either 'count' or 'bytes' may be set, not both")
+
+
+class BatchFormat(enum.Enum):
+    """The format that will be returned from an array read operation."""
+    DENSE = "dense"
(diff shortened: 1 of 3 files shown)


bkmartinjr:
> An open question: For the DENSE type, what are the details of the “coordinates of the slice (e.g. origin, shape) and an Arrow Tensor containing slice values”?

Since the last draft of the abstract spec, my thinking has moved toward simplification, ie,
* don't support dense read/write of sparse NdArray
* don't support sparse reads/writes of dense NdArray
* don't do partial / incremental read of dense NdArray (although George has countered on this one)
In other words, simplify down to the core useful bits.

In the current prototype implementation of NdArray, there is no need to return coordinates for a `dense` read, as the caller already knows them.  If we go with George's proposal, we will need to revisit this.

I think your answer is good - it is a dense sub-array (only), so (`origin`, `shape`, `subarray`) provides all of the necessary info for iterating over partial results.

python-spec/src/somabase/options.py
```diff
@@ -0,0 +1,95 @@
...
+    def _validate(self, attr: attrs.Attribute, value):
+        if not value:
+            return  # None (or 0, which we treat equivalently) is always valid.
+        if value < 0:
+            raise ValueError(f"If set, '{attr.name}' must be positive")
+        if self.count and self.bytes:
+            raise ValueError("Either 'count' or 'bytes' may be set, not both")
+
+
+class BatchFormat(enum.Enum):
```
bkmartinjr, line 76: I primarily defined these in the abstract spec because i was too lazy to define a read/write operation for every data type, and I just wanted to express which formats were expected to be used.

For the Python bindings, it seems more Python to use a specific method name for each type, rather than a generic read method that takes a batchformat parameter.  Should we pick one?

I.e., do we prefer:
> object.read(batch_format="table")

or
> object.read_as_arrow_table()

In the current NdArray prototype, I actually split the difference, and defined:
```
SOMASparseNdArray.read_sparse_tensor(format="coo|csr|csc")  -> pyarrow.SparseTensor subclass
SOMASparseNdArray.write_sparse_tensor(...)

SOMADenseNdArray.read_tensor() -> pyarrow.Tensor
SOMADenseNdArray.write_tensor(tensor)
SOMADenseNdArray.read_numpy() -> numpy.ndarray
SOMADenseNdArray.write_numpy(numpy.ndarray)

SOMASparseNdArray.read_table() -> pyarrow.Table # contains COO data in a Table
SOMASparseNdArray.read_as_pandas() -> pandas.DataFrame  # same as read_table
```

etc (you get the idea)

This seems to be more in keeping with Pandas/Numpy/Arrow API, and provides a clearer API IMHO.

bkmartinjr (approved):
I don't have a lot of experience with the `attrs` class, but it looks reasonable from reading the docs, and I am happy to use it.

My only substantive question on this PR is if we need the BatchFormat type at all.  See inline comment.

python-spec/src/somabase/options.py
johnkerl: re
> object.read(batch_format="table")
vs.
> object.read_as_arrow_table()
i am personally happy either way

johnkerl (approved):

python-spec/src/somabase/options.py
thetorpedodog: OK, I’m glad to read this—I was going to ask about it more when we got to something that implemented that interface. I do think `read_as_x` has the advantage of being a little more common in the Python ecosystem, as well as less error-prone and more useful as far as type checking goes. I’m not a fan of functions which return different overall types depending upon a parameter you pass in; while it is possible to write annotations that cover it (e.g. with `@overload` and the `Literal[some_literal]` syntax), it still leaves the implementor (and potentially the user) with worse tooling feedback and the potential to miss or mismatch entries.

I’ll drop the `BatchFormat` enum for now (with the knowledge that, should we ever need it, it survives in changeset 49cbe30371d8343315c1c71192560452c62bea7f).

johnkerl: me:

> i am personally happy either way

@thetorpedodog :

> I’m not a fan of functions which return different overall types depending upon a parameter you pass in; while it is possible to write annotations that cover it (e.g. with @overload and the Literal[some_literal] syntax), it still leaves the implementor (and potentially the user) with worse tooling feedback and the potential to miss or mismatch entries.

yes, this really tips it -- having specifier args on the input, and `Union[foo, bar]` on the return types -- does indeed muddy the keystroking as well as weakening the value of the type-checker.

so i officially throw in with @thetorpedodog and @bkmartinjr :D

PR #40, opened 2022-10-10
incorporate discussion from 20221007 design review meeting

bkmartinjr:
Changes:
* All offset types (`soma_rowid`, `soma_joinid`) changed from uint64 to positive int64.
* Remove `reshape` operation from the NdArray types.
* Clarify schema type conformance and promotion behavior, particularly with regard to Arrow's `string` and `large_string`
* NdArray COO column names - add `soma` prefix to move all names into the reserved namespace.
* Clarify explicit nature of `soma_rowid` / `soma_joinid` handling throughout.

Optional review:  @ambrosejcarr

### DIFF
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -68,7 +68,7 @@ Numeric index types (eg, offset indexing into dense arrays) are specified with `
 
 SOMA is intended to be strongly typed. With one exception noted below, all requests for a given Arrow type must be fulfilled or generate an error based upon the capabilities of the underlying storage system. Silently casting to a less capable type (eg, float64 to float32) is _not_ permitted. All operations specifying or introspecting the type system must be self-consistent, eg, if object `create` accepts a given Arrow type or schema, the `get schema` operation must return the same types.
 
-SOMA _does_ permit one form of type promotion - variable length types (`sting`, `binary`) may be promoted to their 64-bit variants (`large_string`, `large_binary`) at the time of object creation. However, this promotion must be explicit and visible to the API user via the `get schema` operation.
+SOMA _does_ permit one form of type promotion - variable length types (`string`, `binary`) may be promoted to their 64-bit variants (`large_string`, `large_binary`) at the time of object creation. However, this promotion must be explicit and visible to the API user via the `get schema` operation.
 
 SOMA places no constraints on the underlying types used by the storage system, as long as the API-level representation is consistent across operations, and the supported types full match the Arrow definition of their type semantics.
 
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -310,7 +310,7 @@ read(
 Parameters:
 
 - slices - the rows to read. Defaults to 'all'. Rows are addressable with a row offset (uint), a row-offset range (slice), an Arrow array or chunked array of row-offsets, or a list of both offsets and slices.
-- column_names - the named columns to read and return. Defaults to all. System-defined columns (`soma_rowid` and `soma_joinid`) may be included in this list.
+- column_names - the named columns to read and return. Defaults to all, including system-defined columns (`soma_rowid` and `soma_joinid`).
 - batch_size - a [`SOMABatchSize`](#SOMABatchSize), indicating the size of each "batch" returned by the read iterator. Defaults to `auto`.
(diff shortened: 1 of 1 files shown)


abstract_specification.md
```diff
@@ -296,7 +310,7 @@ read(
 Parameters:
 
 - slices - the rows to read. Defaults to 'all'. Rows are addressable with a row offset (uint), a row-offset range (slice), an Arrow array or chunked array of row-offsets, or a list of both offsets and slices.
-- column_names - the named columns to read and return. Defaults to all. The pseudo-column `soma_rowid` may be included in this list.
+- column_names - the named columns to read and return. Defaults to all. System-defined columns (`soma_rowid` and `soma_joinid`) may be included in this list.
```
johnkerl, line 313: ok sounds like people get back `soma_rowid` and `soma_joinid` on read if they don't specify `column_names` -- correct?

johnkerl (approved):
👍

abstract_specification.md
bkmartinjr: Yes - the logic:
* in a `read`, column_names default to "all", ie, all columns
* for SOMADataFrame, both `soma_rowid` and `soma_joinid` are mandatory columns (only the second is materialized).  For SOMAIndexedDataFrame, `soma_joinid` is mandatory.

So by default all columns, including the `soma_` columns are returned.  But the user can exclude them by using a non-default `column_names` param.

Example:

> SOMADataFrame.read()     # returns all use-defined columns plus `soma_rowid` and `soma_joinid`
> SOMADataFrame.read(column_names=['soma_rowid'])     # returns only `soma_rowid`
> SOMAIndexedDataFrame.read(column_names=['soma_joinid'])     # returns only `soma_joinid`
> SOMAIndexedDataFrame.read(column_names=['foo', 'soma_joinid'])     # returns only columsn `foo` and `soma_joinid`
> SOMAIndexedDataFrame.read(column_names=['soma_rowid'])     # generates error as there is no such column

etc.

thetorpedodog: “may be” sounds vague, as if it’s an implementation decision whether the columns are included. Maybe something like “will include […] if they are matched by the user’s criteria”? This could probably be ironed out further but that’s a place to start.

thetorpedodog (approved):
Overall I think the material of the changes are good. I have one wording suggestion but am OK with the current language if we can’t come up with anything better.

abstract_specification.md
bkmartinjr: how about this as an alternative?
- column_names - the named columns to read and return. Defaults to all, including system-defined columns (`soma_rowid` and `soma_joinid`).

bkmartinjr: @thetorpedodog - sorry, didn't tag you. LMK if the above alternative wording is specific enough.  I agree we need to improve the working, was just trying to wordsmith it a bit more ;-)

abstract_specification.md
```diff
@@ -60,9 +60,17 @@ In the following doc:
...
 
-> ⚠️ **Issue** - the above uses Arrow `string`. Should we be using `large_string` instead (no 2GB cap), or allowing both?
+### Type conformance and promotion
+
+SOMA is intended to be strongly typed. With one exception noted below, all requests for a given Arrow type must be fulfilled or generate an error based upon the capabilities of the underlying storage system. Silently casting to a less capable type (eg, float64 to float32) is _not_ permitted. All operations specifying or introspecting the type system must be self-consistent, eg, if object `create` accepts a given Arrow type or schema, the `get schema` operation must return the same types.
+
+SOMA _does_ permit one form of type promotion - variable length types (`sting`, `binary`) may be promoted to their 64-bit variants (`large_string`, `large_binary`) at the time of object creation. However, this promotion must be explicit and visible to the API user via the `get schema` operation.
```
johnkerl, line 71: nit: `sting` -> `string`

bkmartinjr: fixed - good catch!

gspowley (approved):

abstract_specification.md
thetorpedodog: Your current version sounds good.

Shelnutt2 (approved):

aaronwolen (approved):
LGTM! Thanks for adding these changes.

PR #42, opened 2022-10-12
Rename `type` getter to `soma_type`.

thetorpedodog:
`type` as a name is easy to confuse for (and imo would better fit as) the data type that is stored in an array/dataframe/whatever, rather than just the name of the thing itself. Renaming it `soma_type` will make it clearer that it refers to the SOMA object type, and also allow us to use `type` for a more user-relevant name in the future.


johnkerl (approved):

bkmartinjr (approved):
LGTM

thetorpedodog:
not sure why I didn’t merge this earlier

PR #47, opened 2022-11-04
Add lifecycle stages section

aaronwolen:
Initial draft of the section describing the various lifecycle labels used to annotate SOMA classes and methods. Feedback and suggestions are welcome.

Related to https://github.com/single-cell-data/TileDB-SOMA/issues/487 and [sc-23225].

PR #50, opened 2022-11-18
Refine lifecycle labels

aaronwolen:
Incorporates feedback from the initial PR (#47) and includes example lifecycle annotations for a class, method, and parameter.

Let me know if there are alternative suggestions for the proposed annotation convention.

### DIFF
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -20,12 +20,36 @@ This document is attempting to codify the abstract, language-neutral SOMA data m
 
 # Lifecycle Stages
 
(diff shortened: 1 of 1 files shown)


johnkerl (approved):

thetorpedodog (approved):
This is a little verbose (I might go with just a `[maturing]` tag or something, or maybe `[E]`/`[M]`/`[S]`/`[P]`) but this is fine to start with; we can revise it if need be.

(As I have mentioned before, I am also considering creating/using decorators in the style of [deprecated](https://pypi.org/project/Deprecated/) to give us a more structured version of this.)

PR #54, opened 2022-12-13
Describe `context` field and separate out common operations.

thetorpedodog:
- Adds a description of the `context` field and how SOMA implementations use context data.
- Separates out the `context` and `metadata` operations into a new "common operations" section.


abstract_specification.md
```diff
@@ -649,6 +655,12 @@ Examples, using a pseudo-syntax:
 
 ## Platform-Specific Configuration
 
+SOMA includes provisions for two separate ways to provide platform-specific configuration data to objects and operations.
+For configuration settings that affect a single operation, SOMA functions include a `platform_config` parameter, which can be used on a call-by-call basis to specify short-term behavior.
```
bkmartinjr, line 659: I find the `short-term behavior` phrase confusing, as some of the `platform_config` parameters modify schema, which has long-term effects.

Suggest using `... to specify per-operation behavior`, or even more explicitly, `... to specify behavior of a single operation call.`

abstract_specification.md
```diff
@@ -649,6 +655,12 @@ Examples, using a pseudo-syntax:
...
+For long-lived configuration settings, SOMA objects expose a `context` field that contains implementation-defined data that can be shared over an entire analysis session.
```
bkmartinjr, line 660: Perhaps: "For multi-operational configuration settings, ..."?

abstract_specification.md
```diff
@@ -685,14 +695,30 @@ The exact contents of each platform’s entry in the configuration mapping are f
...
+#### Implementation and usage guidelines
 
 The configuration passed in the `platform_config` object is intended for operation-specific tuning (though it may make sense for user code to use the same `platform_config` across multiple operations). A `platform_config` should only be used for the operation it was provided to (and to directly dependent operations); it should not be stored for later calls. Environmental configuration (like login credentials or backend storage region) should be provided via the (to-be-defined) Context object.
 
 Operations should not *require* a `platform_config` entry to complete; the platform should use a sensible default if a given configuration value is not provided. Required environmental data generally belongs in the Context object.
 
 A platform should only ever examine its own entry in the `platform_config` mapping. Since the structure and contents of each entry is wholly implementation-defined, one platform cannot make any assumptions at all about another’s configuration, and for predictability should avoid even trying to extract any information from any other configuration entries.
 
+### Long-lived context data
```
bkmartinjr, line 706: A point of confusion:   as I read this, the `context` field is read-only, has no user-accessible operations, and is implementation specific.  So why specify it - why not treat it as hidden state?  How does the user "set" the context?

I think the confusion is that the user _can_ specify a context in some operations (eg, `open`).  So perhaps change the first sentence to:  "... described above, the user can specify a `context` value in certain operations.  The context value is an implementation-dependent, multi-operation configuration object, which will provide session-like state."

Do we also need to specify how it is inherited/passed, and the lifespan?  The text says "global state", which seems to imply I can only have one context?  But I suspect that there are operations where I can "set" the context, and others that will inherit it.

Lastly: do we need the concept of a default context?  I don't think we do, but it might simplify the description of how the context propagates.

bkmartinjr (changes requested):
Mostly nits, but a bit of confusion around the context definition.

abstract_specification.md
```diff
@@ -685,14 +695,30 @@ The exact contents of each platform’s entry in the configuration mapping are f
...
+Examples of data that might belong in a context includes:
+
+- Storage credentials or API keys
+- Endpoint URLs
+- Database connections
+
+In other words, the `context` contains what is effectively the global state for a SOMA analysis session.
+
+The format and contents of a `context` object is completely implementation-defined.
+An implentation may provide a way to construct its specific context type, which can be used by user setup code to connect to the SOMA data store.
```
atolopko-czi, line 719:
```suggestion
An implementation may provide a way to construct its specific context type, which can be used by user setup code to connect to the SOMA data store.
```

abstract_specification.md
thetorpedodog: I've tried to rephrase this.

abstract_specification.md
thetorpedodog: Reworked this along with the below.

thetorpedodog (commented):
Our new object lifecycle (`Type.open` and `Type.create` as class-level methods; `.add_new_X` in collections) has made the way Context works a lot clearer (I think). We do still need to update the spec to reflect this new lifecycle (and the rest of everything, like open/close, and read/write) but that is a separate task.

I think this is ready for another look.

thetorpedodog:
> I think this is ready for another look.

or it would have been, had I remembered to commit and push. (which I had not at that point, but have now.)

thetorpedodog:
> We do still need to update the spec to reflect this new lifecycle (and the rest of everything, like open/close, and read/write) but that is a separate task.

Which, for reference, is https://github.com/single-cell-data/SOMA/issues/107.

abstract_specification.md
bkmartinjr: It still begs the question of what an "analysis session" is, and how it relates to the lifespan of the object.  Can this be tightened to something like "... shared over the life of the object?"

abstract_specification.md
bkmartinjr: Are these changes universal, to all implementations?  Or are they Python ABI specific?   If only the latter (which seems quite possible), then they probably don't belong in this spec.

bkmartinjr: with the example, I think this is sufficiently clear (for now).

bkmartinjr (approved):

abstract_specification.md
thetorpedodog: It can last beyond the lifecycle of any single object (cf. the manual-context-passing part of the example), so I'm not sure "the life of the object" quite captures it; I'll give this a little time to stew.

johnkerl (approved):

aaronwolen (approved):

abstract_specification.md
```diff
@@ -650,6 +656,12 @@ Examples, using a pseudo-syntax:
 
 ## Platform-Specific Configuration
 
+SOMA includes provisions for two separate ways to provide platform-specific configuration data to objects and operations.
+For configuration settings that affect a single operation, SOMA functions include a `platform_config` parameter, which can be used on a call-by-call basis to change implementation-specific behavior for that call.
+For configuration settings whose scope affects many operations, SOMA objects expose a `context` field that contains implementation-defined data that can be shared over an entire analysis session.
```
atolopko-czi, line 661: We might want to also mention the parent/child inheritance of the context here, to introduce the idea before it is detailed below. Something like the below. I ended up editing both lines (platform_config and context), for consistency.

```suggestion
All SOMA operations include a `platform_config` parameter that allow a caller to provide implementation-specific configuration settings that affect the behavior of an operation on per-call basis.
All SOMA objects expose a `context` field that contain implementation-specific configuration settings that affect all operations called on a SOMA object and any of its children, and that are scoped to the life of these objects.
```

atolopko-czi (approved):
LGTM as is, but I added a few suggestions for your consideration.

abstract_specification.md
thetorpedodog: This is a good idea. It should also become clearer when we specify the object lifecycle methods in a soon-to-come change.

abstract_specification.md
```diff
@@ -686,14 +696,64 @@ The exact contents of each platform’s entry in the configuration mapping are f
...
 
 Operations should not *require* a `platform_config` entry to complete; the platform should use a sensible default if a given configuration value is not provided. Required environmental data generally belongs in the Context object.
 
 A platform should only ever examine its own entry in the `platform_config` mapping. Since the structure and contents of each entry is wholly implementation-defined, one platform cannot make any assumptions at all about another’s configuration, and for predictability should avoid even trying to extract any information from any other configuration entries.
 
+### Long-lived context data
+
+In addition to the `platform_config` parameters described above, implementations can also store a value on the `context` field of a SOMA object.
+This read-only field should be used to store long-lived implementation-specific data that is used to access SOMA datasets.
```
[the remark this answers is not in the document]
thetorpedodog: Extremely good idea.

abstract_specification.md
thetorpedodog: I also added some language here to note that context can be shared across more than just a single parent/child set.

PR #56, opened 2022-12-14
python: Build out base data storage ABCs.

thetorpedodog:
This adds metadata to the base SOMA objects and specifies the major read/write operations of the data storage classes: the NDArrays and DataFrame.

Lifecycle management (opening, closing, etc.) will be coming in future changes.

### DIFF
--- a/python-spec/src/somabase/data.py
+++ b/python-spec/src/somabase/data.py
@@ -113,7 +113,7 @@ class DenseNDArray(NDArray, metaclass=abc.ABCMeta):
         *,
         batch_size: options.BatchSize = options.BatchSize(),
         partitions: Optional[options.ReadPartitions] = None,
-        result_order: options.ResultOrder = None,  # XXX: Set default
+        result_order: options.ResultOrder = None,  # TODO: Set default
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> pyarrow.Tensor:
         """Reads the specified subarray from this NDArray as a Tensor.
@@ -152,7 +152,7 @@ class SparseNDArray(NDArray, metaclass=abc.ABCMeta):
         *,
         batch_size: options.BatchSize = options.BatchSize(),
         partitions: Optional[options.ReadPartitions] = None,
-        result_order: options.ResultOrder = None,  # XXX set default
+        result_order: options.ResultOrder = None,  # TODO: Set default
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> "SparseRead":
         """Reads a subset of the object in one or more batches.
@@ -206,40 +206,42 @@ _T = TypeVar("_T")
 class ReadIter(Iterator[_T], metaclass=abc.ABCMeta):
     """SparseRead result iterator allowing users to flatten the iteration."""
 
-    # __iter__ is already implemented as `return self` in Iterator
+    # __iter__ is already implemented as `return self` in Iterator.
     # SOMA implementations must implement __next__.
 
-    # XXX: Considering the name "flat" here too.
     @abc.abstractmethod
-    def all(self) -> _T:
-        """Returns all the requested data in a single operation."""
+    def concat(self) -> _T:
+        """Returns all the requested data in a single operation.
+
+        If some data has already been retrieved using ``next``, this will return
+        the rest of the data after that is already returned.
+        """
         raise NotImplementedError()
 
 
-class SparseRead(metaclass=abc.ABCMeta):
-    """Intermediate type to allow users to format when reading a sparse array."""
+class SparseRead:
+    """Intermediate type to choose result format when reading a sparse array.
 
-    @abc.abstractmethod
+    A query may not be able to return all of these formats. The concrete result
+    may raise a ``NotImplementedError`` or may choose to raise a different
+    exception (likely a ``TypeError``) containing more specific information
+    about why the given format is not supported.
+    """
+
     def coos(self) -> ReadIter[pyarrow.SparseCOOTensor]:
         raise NotImplementedError()
 
-    @abc.abstractmethod
     def cscs(self) -> ReadIter[pyarrow.SparseCSCMatrix]:
         raise NotImplementedError()
 
-    @abc.abstractmethod
     def csrs(self) -> ReadIter[pyarrow.SparseCSRMatrix]:
         raise NotImplementedError()
 
-    # XXX Does this need to return a ReadIter of Tuple[coordinates, Tensor]?
-    @abc.abstractmethod
     def dense_tensors(self) -> ReadIter[pyarrow.Tensor]:
         raise NotImplementedError()
 
-    @abc.abstractmethod
     def record_batches(self) -> ReadIter[pyarrow.RecordBatch]:
         raise NotImplementedError()
 
-    @abc.abstractmethod
     def tables(self) -> ReadIter[pyarrow.Table]:
         raise NotImplementedError()
--- a/python-spec/src/somabase/data.py
+++ b/python-spec/src/somabase/data.py
@@ -5,12 +5,14 @@ members will be exported to the ``somabase`` namespace.
 """
 
 import abc
-from typing import Tuple
+from typing import Any, Iterable, Optional, Sequence, Tuple, Union
 
+import attrs
 import pyarrow
 from typing_extensions import Final
 
 from somabase import base
+from somabase import options
 
 
 class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
@@ -18,7 +20,31 @@ class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
 
     __slots__ = ()
 
-    # TODO: Read/write.
+    # Data operations
+
+    @abc.abstractmethod
+    def read(
+        self,
+        ids,  # TODO: Specify type
+        column_names: Optional[Sequence[str]] = None,
+        *,
+        batch_size: options.BatchSize = options.BatchSize(),
+        partitions: Optional[options.ReadPartitions] = None,
+        result_order: Optional[options.ResultOrder] = None,
+        value_filter: Optional[str] = None,
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> Iterable[pyarrow.Table]:
+        """Reads a user-defined slice of data into Arrow tables."""
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def write(
+        self,
+        values: Union[pyarrow.RecordBatch, pyarrow.Table],
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> None:
+        """Writes values to the data store."""
+        raise NotImplementedError()
 
     # Metadata operations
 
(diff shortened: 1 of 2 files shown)


python-spec/src/somabase/data.py
```diff
@@ -0,0 +1,220 @@
...
+
+class NDArray(base.SOMAObject, metaclass=abc.ABCMeta):
+    """Common behaviors of N-dimensional arrays of a single primitive type."""
+
+    __slots__ = ()
+
+    # Data operations (listed alphabetically)
+
+    @abc.abstractmethod
+    def read_coo(
```
bkmartinjr, line 89: Couple of issues here:
* soma.DenseNDArray read operation can't return an iterator, as it only returns a well-defined slice, whereas SparseNDArray _does_ return an iterator.
* ideally, the naming scheme would be normalized to allow for a variety of ecosystem/data model returned, as well as "iterator" or "no iterator"

`read_coo` doesn't tell me if this is an Arrow Tensor, Scipy COO spmatrix, COO data in a Table, etc.  It is probably fine if "Arrow" is the default ecosystem and implied, but I'd like to define the other typical cases.

Maybe something like:  <op>_all_"as"_<ecosystem>, where _as_ecosystem and _all are optional, and:
* Arrow is implied if ecosystem not stated stated
* For NDArray and DataFrame, Iterator is implied if no _all

So, examples for SparseNDArray:
* read_coo - iterator of Arrow Sparse Tensor
* read_table - iterator of Arrow Table
* read_dense - iterator of Arrow dense tensor
* read_as_scipy - iterator of scipy.sparse.coo_array (or coo_matrix)
* read_as_numpy - iterator of densified Numpy NdArray
* read_table_all - flatten output of read_table
* read_coo_all - flatten output of read_coo
etc.

DenseNDArray would be slightly different as it is currently not iterable, and would support:
* read_dense - Arrow Tensor (_not_ an iterator)
* read_dense_as_numpy - NDArray (_not_ an iterator)
etc.

Thoughts?

bkmartinjr:
Tagging a few folks who we should get comments from: @mlin @aaronwolen @atolopko-czi

thetorpedodog:
Before I address anything in specific, I want to split this up into two separate stages:

1. #58 (only the basics)
2. This part

That way we can get work done that depends upon the existence of these interfaces without requiring them to be fully specified.

python-spec/src/somabase/data.py
thetorpedodog: This could end up being a huge number of methods to attach to each of these classes. Would it make sense to have something like:

```python
class SparseNDArray(...):
  # ...
  def read(self, ...) -> "SparseResult":
    ...


class SparseResult:
  # ...
  def as_coos(self) -> "ResultIterable[arrow.Tensor]":
    ...

  def as_tables(self) -> "ResultIterable[arrow.Table]":
    ...

class ResultIterable(Iterable[_T]):
  # Iterable methods for _T

  def flatten(self) -> _T:
    """Flattens all the results from this into one value."""
  # Alternately: .all
```

Then, using it would look something like this:

```python
my_table = some_nd_array.read(...).as_tables().flatten()
# or
for partial in some_nd_array.read(...).as_scipys():
  ...
```

These are just some first ideas for names, but I think doing something like this would be useful to separate the querying/filtering/loading results etc. part from the converting results to the desired format part (and make the method set of NDArray and friends much less intimidating).

I think this also means we could implement `ResultIterable` for each type as reusable wrappers around `Iterable`, since no matter what backend you’re, flattening a sequence of arrow `Table`s together ends up being the same thing:

```python

class Tables(ResultIterable[arrow.Table]):

  # __next__ and __iter__ forward to wrapped iterable

  def flatten(self) -> arrow.Table:
    # the actual code which actually flattens arrow Tables
    # which I am too lazy to look up what function it is
   ...
```

bkmartinjr: I am definitely open to this idea!  At first glance, it appears cleaner from a user vantage (eg, the patterns learned with one type apply to others).  I'm not sure there are many places where we get a shared implementation, but that does not diminish the idea at all.

If we move to a API that supports explicit use of \_\_getitem__ (as proposed in single-cell-data/TileDB-SOMA#573), how would these be integrated together?  Or are they exclusive?

thetorpedodog:
After longer than I had hoped to, here is a more fleshed-out version of the proposal I discussed earlier.

The basic structure looks something like this:

```python
some_dense_array.read(...).tables()
# -> an iterator of Arrow Tables
some_dense_array.read(...).csrs().all()
# -> a single flattened sparse CSR matrix
```

After some more implementation I generally like how it looks overall (though I am not overly attached to it). If we do like this, a few details I have been thinking about include:

  - Should the various types be functions or properties?
    ```python
    some_array.read(...).tables
    # vs.
    some_array.read(...).tables(...)
    ```
    Property syntax is less syntax to type out. Function syntax makes it seem a little more like a transform (which to some extent it is) and allow for if we need to put some kind of options in there in the future.

  - Singular or plural (e.g. `.table` vs `.tables`)? I have been thinking plural because, while it is a bit longer, it doesn’t risk the confusion of expecting `.table()` to return a single table.

  - To `as_` or not to `as_`. I don’t think it adds much in this case.

  - The naming of the `all` method. Other options I thought of include `flat` and `flatten`, but there are certainly more good ideas.

There is no doubt some stuff that I did not think of so get at me with all your feedback.

johnkerl:
@thetorpedodog this is baller!! :) 

I personally lean more toward `.tables()` with the explicit `()` -- makes it clear it's computing something rather than just accessing something that is already computed

python-spec/src/somabase/data.py
```diff
@@ -100,3 +191,55 @@ def nnz(self) -> int:
...
+
+
+class SparseRead(metaclass=abc.ABCMeta):
+    """Intermediate type to allow users to format when reading a sparse array."""
+
+    @abc.abstractmethod
+    def coos(self) -> ReadIter[pyarrow.SparseCOOTensor]:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
```
bkmartinjr, line 226: these only make sense if it is a 2D array (matrix).  What is the expected behavior if it is not 2D?  Raise?

bkmartinjr:
On your questions:

> Should the various types be functions or properties?

I slightly prefer function. But not enough to argue about it :-)

> Singular or plural

plural, again, slight preference

> `all`

Agree that `.flat()` (or my favorite, `concat()`, borrowed from pyarrow) is much better than `all`.

bkmartinjr:
Additional questions:

1. We also have support for ecosystem to/from hooks, eg., "read dataframe and return result as Pandas".  Is the intent of this proposal to remove that (and rely solely on Arrow for conversion)?  Or do we need to include hooks for that as well?   I believe most (if not all) of the current code is redundant with an Arrow method (eg, soma.DataFrame.read_as_pandas() is essentially identical to soma.DataFrame.read().to_pandas().

2. I'm trying to think about how this style manifests in the soma.Experiment.query() API, where the query() "result" is a deferred query object that has accessor methods for the named slots ("X", "obs", etc).  I think the right answer is that each of those "slots" returns the same pattern as their underlying type, eg.,
    ```python
    with experiment.query(...) as query:
        the_entire_obs_slice = query.obs().tables().concat()
    ````
    or, if you want it in Pandas, use the Arrow Table.to_pandas() function:
    ```python
    with experiment.query(...) as query:
        the_entire_obs_slice = query.obs().tables().concat().to_pandas()
    ````

Thoughts?

python-spec/src/somabase/data.py
thetorpedodog: Raising an error seems like a good idea. Initially this can be a basic `ValueError` but in the future we can build out some more granular exception types to allow better error handling.

thetorpedodog:
> Additional questions:
> 
> 1. We also have support for ecosystem to/from hooks, eg., "read dataframe and return result as Pandas".  Is the intent of this proposal to remove that (and rely solely on Arrow for conversion)?  Or do we need to include hooks for that as well?   I believe most (if not all) of the current code is redundant with an Arrow method (eg, soma.DataFrame.read_as_pandas() is essentially identical to soma.DataFrame.read().to_pandas().

My thought would be to just use Arrow for conversions here, so that there is one obvious way to do it right. We could consider adding other 

> 2. I'm trying to think about how this style manifests in the soma.Experiment.query() API, where the query() "result" is a deferred query object that has accessor methods for the named slots ("X", "obs", etc).  I think the right answer is that each of those "slots" returns the same pattern as their underlying type, eg.,
>    ```python
>    with experiment.query(...) as query:
>        the_entire_obs_slice = query.obs().tables().concat()
>    ```
>      
>    or, if you want it in Pandas, use the Arrow Table.to_pandas() function:
>    ```python
>    with experiment.query(...) as query:
>        the_entire_obs_slice = query.obs().tables().concat().to_pandas()
>    ```
> 
> Thoughts?

I like this idea; reusing the pattern will make it consistent and hopefully straightforward to implement.

python-spec/src/somabase/data.py
bkmartinjr: Sounds good - so the pattern is to have  consistent set of accessor method names, and just error out those that don't make sense. The only downside to this will be documentation - i.e., there may be methods that are nonsensical on a DataFrame but work fine on an NDArray.

bkmartinjr:
> My thought would be to just use Arrow for conversions here, so that there is one obvious way to do it right. We could consider adding other

I also prefer this. Our routines are almost always just a wrapper on the same underlying function....

And re: query
> I like this idea; reusing the pattern will make it consistent and hopefully straightforward to implement.

Good - I'll give it a try in PR single-cell-data/TileDB-SOMA#632 and see what shakes out.

python-spec/src/somabase/data.py
```diff
@@ -100,3 +191,55 @@ def nnz(self) -> int:
...
+# Sparse reads are returned as an iterable structure:
+
+
+class ReadIter(Iterator[_T], metaclass=abc.ABCMeta):
+    """SparseRead result iterator allowing users to flatten the iteration."""
+
+    # __iter__ is already implemented as `return self` in Iterator
+    # SOMA implementations must implement __next__.
+
+    # XXX: Considering the name "flat" here too.
```
bkmartinjr, line 212: `flat` or `concat` seem better to me, as they more clearly indicate what happens if the __next__ has already been called on the iterator....

python-spec/src/somabase/data.py
bkmartinjr: BTW, given this approach, perhaps this these methods should not be abstract -- i.e., subclasses should inherit the default implementation (raising the "NotImplemented" exception when they are not implemented).

python-spec/src/somabase/data.py
thetorpedodog: changed to `concat` (while `flat` is a bit more fun, consistency with Arrow is more useful)

python-spec/src/somabase/data.py
thetorpedodog: un-abstracted these.

thetorpedodog:
This, I believe, should address the comments from earlier versions. Still some longer-term TODOs to iron out but those are suitable for follow-ups.

bkmartinjr (approved):

johnkerl (approved):
🚀

PR #57, opened 2022-12-14
Add pre-commit rules for the Python implementation.

thetorpedodog:
pre-commit allows an author to easily verify that their code follows the rules (formatting etc.) before committing it.  This initial version does a simple formatting check in the Python directory (at this point, primarily to prevent me from committing formatting mistakes). We can add more rules later.

To use pre-commit, first install the pre-commit program (available in apt on Debian, or from https://pre-commit.com/) and run `pre-commit install` in the repository.

This change also brings existing files into compliance with the current pre-commit ruleset.


bkmartinjr (approved):

johnkerl (commented):
🚢

PR #58, opened 2022-12-15
python: Add basis of data storage types

thetorpedodog:
Split off from #56. This adds the data storage types (DataFrame, the NDArrays), but does not include the more complex read/write parts of the API. This will allow us to build things that refer to these types without waiting for the exact details of how reading and writing are implemented.


bkmartinjr (approved):
LGTM

PR #60, opened 2022-12-17
Add tools to create simple composed types and implement Measurement

thetorpedodog:
This change implements the Measurement "behavior class" in two steps:

 1. Creates a collection proxy that can be used as a base class to decorate Collections with additional behaviors.
      - Also a descriptor object that makes writing attribute-to-indexing bridges trivial.
 2. Implements Measurement using the proxy and the descriptors.

From here, implementing the other composed classes is quite straightforward. Once all are complete, TileDB-SOMA will no longer need to use its own custom implementation of Experiment and Measurement and can instead use the common wrapper classes around its own Collection objects.

---

This is the first part of the Grand Plan to separate behavior from storage. I’m sure there are parts of it that I have not explained quite as well as I had hoped so I am happy for questions or feedback on the direction as a whole in addition to the specifics of this code.

### DIFF
--- a/python-spec/src/somabase/_wrap.py
+++ b/python-spec/src/somabase/_wrap.py
@@ -1,0 +1,220 @@
+from typing import (
+    Any,
+    Generic,
+    ItemsView,
+    Iterator,
+    KeysView,
+    MutableMapping,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    ValuesView,
+    overload,
+)
+
+import attrs
+from typing_extensions import Final
+
+from somabase import base
+
+_ST = TypeVar("_ST", bound=base.SOMAObject)
+_T = TypeVar("_T")
+
+_SENTINEL = object()
+
+
+class CollectionProxy(base.Collection[_ST]):
+    """Base class to forward SOMA collection methods to a "real" SOMA object.
+
+    This is intended for use as a mixin for SOMA objects that decorate standard
+    collection types with behaviors, e.g. the Experiment or Measurement classes.
+    It allows the separation of the behaviors (implemented in the decorator)
+    from the backend implementation (implemented in the wrapped object).
+    """
+
+    __slots__ = ("_backing",)
+
+    def __init__(self, backing: base.Collection[_ST]):
+        """Creates a new CollectionProxy backed by the given collection."""
+        self._backing: Final = backing
+        """The object that actually provides the indexing for this object."""
+
+    # SOMA methods
+
+    @property
+    def metadata(self) -> MutableMapping[str, str]:
+        return self._backing.metadata
+
+    def unwrap(self) -> base.Collection[_ST]:
+        """Unwrap gets the actual collection backed by this proxy."""
+        inst: base.Collection[_ST] = self
+        while isinstance(inst, CollectionProxy):
+            inst = inst._backing
+        return inst
+
+    # Sized
+
+    def __len__(self) -> int:
+        return len(self._backing)
+
+    # Iterable
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self._backing)
+
+    # Container
+
+    def __contains__(self, key: Any) -> bool:
+        return key in self._backing
+
+    # Mapping
+
+    def __getitem__(self, key: str) -> _ST:
+        return self._backing[key]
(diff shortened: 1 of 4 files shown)


python-spec/src/somabase/_wrap.py
```diff
@@ -0,0 +1,220 @@
...
+        default: _ST = None,  # type: ignore[assignment]
+    ) -> _ST:
+        return self._backing.setdefault(key, default)  # type: ignore[arg-type]
+
+
+@attrs.define()
+class item(Generic[_T]):
+    """Descriptor to transform property access into indexing.
+
+    This descriptor works with :class:`CollectionWrapper` to allow for simple
```
bkmartinjr, line 139: I could not find a CollectionWrapper - is this supposed to be `CollectionProxy`?

thetorpedodog: Whoops! Fixed.

python-spec/src/somabase/_wrap.py
```diff
@@ -0,0 +1,220 @@
...
+        inst.second = 500
+    """
+
+    typ: Optional[Type[_T]] = None
+    """The type we expect to return from this field."""
+
+    item_name: Optional[str] = None
+    """The name of the item we are getting (``x._backing["whatever"]``).
+
+    This uses the name of the field by default but can be manually overridden.
```
bkmartinjr, line 162: Is there a use case in mind to motivate having both `item_name` and `field_name`?  (I couldn't find any code that uses it, and am trying to understand the reason to have both)

python-spec/src/somabase/_wrap.py
```diff
@@ -0,0 +1,220 @@
...
+        assert self.item_name is not None
+        try:
+            # TODO: Type-check params/returns?
+            return inst[self.item_name]
+        except KeyError as ke:
+            raise AttributeError(
+                f"{_typename(inst)!r} object has no attribute {self.field_name!r}"
+            ) from ke
+
+    def __set__(self, inst: CollectionProxy, value: _T) -> None:
```
bkmartinjr, line 197: Non-blocking question: down the road, if we need to implement additional constraint checks upon __set__ (e.g., perhaps check the shape of the value), how would this be extended or composed by the parent class (e.g., Measurement)?

python-spec/src/somabase/_wrap.py
bkmartinjr: NM....I understand after reading the tests!

bkmartinjr (approved):
One question below about how to extend constraints, in the future, beyond typing.  But this doesn't block approval in my view.

Thanks @thetorpedodog

johnkerl (approved):

python-spec/src/somabase/_wrap.py
thetorpedodog: I think the way to do this would be to add a function similar to the way `@property` (itself a descriptor) works. This would look something like this:

```python

class item(...):
  # ...

  def validator(self, validator_func: Callable[...]):
    self._validator = validator
    return validator

  def __set__(self, inst, value):
    try:
      self._validator(value)
    except ...:
      raise ...
    # existing __set__ implementation
```

which would then be used like this:

```python
class SomeWrapper(_wrap.CollectionProxy):

  it = _wrap.item()

  @it.validator
  def _validate_it(self, new_it):
    if new_it.name in {"john", "paul", "george", "ringo"}:
      raise ValueError("no beatles allowed")
```

This might also be implemented as some kind of transform, where the value gets passed through a `it.transform_set` function or the like:

```python
  it = _wrap.item()

  @it.transform_set
  def _set_it(self, new_it):
    if new_it.name in ...:
      ...  # validation occurs as part of the transform step
    return do_something_to(new_it)
```

If an implementation needs to do its own custom validation or transformation, the value works just like any other property:

```python
class SpecialWrapper(some_mod.SomeWrapper):

  @property
  def it(self):
    return super().it

  @it.setter
  def _set_it(self, new_it):
    new_new_it = special_xform(new_it)
    super().it = new_new_it
```

bkmartinjr: Great - I was thinking along similar lines.  Many thanks.

aaronwolen (approved):

PR #61, opened 2023-01-03
Add SOMADataFrame count accessor

johnkerl:
One of the first questions users ask about a single-cell dataset is "how many cells and how many genes are in here?"

We need to let people answer this question.

This is essential for uptake UX.

### DIFF
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -284,6 +284,7 @@ Parameters:
 | get soma_type                           | Returns the constant "SOMADataFrame"                                           |
 | get schema -> Arrow.Schema              | Return data schema, in the form of an Arrow Schema                             |
 | get index_column_names -> [string, ...] | Return index (dimension) column names.                                         |
+| get count -> int                        | Return the number of rows in the SOMADataFrame.                                |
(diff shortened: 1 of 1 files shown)


bkmartinjr (approved):

abstract_specification.md
```diff
@@ -284,6 +284,7 @@ Parameters:
 | get soma_type                           | Returns the constant "SOMADataFrame"                                           |
 | get schema -> Arrow.Schema              | Return data schema, in the form of an Arrow Schema                             |
 | get index_column_names -> [string, ...] | Return index (dimension) column names.                                         |
+| get count -> int                        | Return the number of rows in the SOMADataFrame.                                |
```
thetorpedodog, line 287: This probably makes sense to be implemented as `__len__` in Python (in addition to the explicit `count` accessor).

thetorpedodog (approved):

abstract_specification.md
bkmartinjr: absolutely!

johnkerl:
(Merging since @aaronwolen has approved https://github.com/single-cell-data/TileDB-SOMA/pull/637)

PR #64, opened 2023-01-06
Update licensing information for PyPI publishing.

thetorpedodog:
Adds TileDB to the copyright file and sets MIT license metadata.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (c) 2021-2023 The Chan Zuckerberg Initiative Foundation
+Copyright (c) 2022-2023 TileDB, Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/python-spec/pyproject.toml
+++ b/python-spec/pyproject.toml
@@ -9,6 +9,8 @@ dynamic = ["version"]
 readme = "./README.md"
 dependencies = ["attrs>=22.1", "numpy>=1.21", "pyarrow", "typing-extensions"]
 requires-python = "~=3.7"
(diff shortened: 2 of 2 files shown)


LICENSE
```diff
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (c) 2021 Ambrose J Carr
```
bkmartinjr, line 3: At a minimum, Ambrose should be assigning his copyright to CZI (not claiming it personally).  I believe we need to add "2022 Chan Zuckerberg Initiative Foundation" as well.

@ambrosejcarr - can you illuminate the personal copyright?  Pretty sure that is counter to your employment agreement :-)

ambrosejcarr: I think this is an artifact of github auto-creation templates, and is not intentional. 

I agree to assign my copyright to CZI or consider another entity recommended by this group.

thetorpedodog: Could you be the one to make that change formally in the file itself? It seems like it would be unwise for me to be the one to edit that line.

ambrosejcarr: https://github.com/single-cell-data/SOMA/pull/77

python-spec/pyproject.toml
```diff
@@ -9,6 +9,8 @@ dynamic = ["version"]
 readme = "./README.md"
```
bkmartinjr, line 9: Once this package is available, it will likely get confused with `tiledbsoma`.  I am wondering if we need the README (or something) to point people toward that repo/package?

bkmartinjr (approved):

python-spec/pyproject.toml
thetorpedodog: A good idea for a followup.

PR #66, opened 2023-01-07
Fix: Make SparseNDArray take sparse arrow types.

thetorpedodog:


python-spec/src/somacore/data.py
```diff
@@ -140,6 +140,15 @@ def write(
     soma_type: Final = "SOMADenseNDArray"
 
 
+SparseArrowData = Union[
+    pyarrow.SparseCSCMatrix,
+    pyarrow.SparseCSRMatrix,
+    pyarrow.SparseCOOTensor,
+    # TODO: add pyarrow.Table?
+]
+"""Any of the sparse data storages provided by Arrow."""
```
bkmartinjr, line 149: +1 for Table.  We support it on read and have naming conventions for the dims/data.  It is also one less copy for use with TileDB (tensors are not yet zero copy)

bkmartinjr (approved):
looks good!  would also like to see Table added to sparse write union.

python-spec/src/somacore/data.py
thetorpedodog: Added.

PR #68, opened 2023-01-09
Set up project versioning for PyPI upload.

thetorpedodog:
Because our project is not at the root of our Git repository, we can't use straight `setuptools_scm`, so we do a little extra work.

- When building a package for release, run `write-version-file` before running `build`/`twine`.
- After that, everything works like normal.

---

A slightly different solution than we use in `TileDB-SOMA` but lets us have a purely declarative `pyproject.toml`, with just a bit of extra work in the build process.

### DIFF
--- a/python-spec/.gitignore
+++ b/python-spec/.gitignore
@@ -1,2 +1,4 @@
 **/__pycache__
 **/*.egg-info
+dist
+src/somacore/_generated_version.py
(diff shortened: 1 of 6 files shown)


johnkerl (approved):

bkmartinjr (approved):

thetorpedodog:
@mlin this has a slightly different set of constraints than the way you do versioning on `TileDB-SOMA` but you might be interested in this solution

mlin:
@thetorpedodog LGTM thanks; I have no particular attachment to the [ad hoc version.py script](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/version.py#L4) used in TileDB-SOMA, I just switched to something I knew works ok after reading the setuptools_scm issues threads. The script does work reasonably well in the local checkout case (i.e. gives you a meaningful version number incl. '-dirty' if applicable, by looking at git status).

PR #69, opened 2023-01-09
Create a build/releaser for the Python somacore library.

thetorpedodog:


.github/workflows/python-somacore.yaml
```diff
@@ -0,0 +1,69 @@
...
+  release:
+    types: [published]
+
+jobs:
+  format-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+      - uses: psf/black@stable
```
bkmartinjr, line 17: These GH Action integrations are sweet.

@johnkerl - what do you think about doing the same in tiledbsoma?

bkmartinjr (approved):

.github/workflows/python-somacore.yaml
johnkerl: SIx of one, half a dozen of the other as far as I am concerned

The brevity of syntax here is nice

On the other hand, there are people whose opinions I value who are strongly in favor of a concise, transparent script that's easier to debug (when there will inevitably be an occasion) than more 'opaque' canned GH actions

I'm not against doing this in TileDB-SOMA -- in fact on net, I'm (slightly) more in favor than against -- just acknowledging multiple points of view ...

PR #72, opened 2023-01-09
Add `py.typed` marker file for downstream type checking.

thetorpedodog:
In order for mypy to use the types in an installed package's code for type-checking purposes, it needs this marker file.  See, for example, https://github.com/python/mypy/blob/master/mypy/py.typed


johnkerl (approved):
🚀

PR #74, opened 2023-01-11
Improve types for coordinate and result-order params.

thetorpedodog:
- Imports the types for each coordinate from TileDB-SOMA.
- Creates a `ResultOrder` enum (which can also be passed as a str).


python-spec/src/somacore/options.py
```diff
@@ -85,10 +89,56 @@ def _validate(self, attr: attrs.Attribute, value):
...
-ReadCoords = Any
-"""Type alias for specifying coordinates for reads. TODO: Define me."""
+    AUTO = "auto"
+    ROW_MAJOR = "row-major"
+    COLUMN_MAJOR = "column-major"
+
+
+DenseCoord = Union[int, slice]
+"""A single coordinate range for reading dense data."""
+DenseNDCoords = Sequence[DenseCoord]
```
thetorpedodog, line 107: This used to be a `Tuple[DenseCoordinates, ...]`; I made it a sequence here since it doesn’t look like we’re using the tupleness of it at all.

python-spec/pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "somacore"
 description = "Python-language API specification and base utilities for implementation of the SOMA system."
 dynamic = ["version"]
 readme = "./README.md"
-dependencies = ["attrs>=22.1", "pyarrow", "typing-extensions"]
+dependencies = ["attrs>=22.1", "numpy", "pyarrow", "typing-extensions"]
```
bkmartinjr, line 10: I think you need `numpy>=1.21` as some of the `numpy.typing` stuff was introduced in that release (eg., `numpy.typing.NDArray`)

bkmartinjr (approved):
Other than the missing dependency version on numpy, this LGTM

johnkerl (approved):

python-spec/pyproject.toml
thetorpedodog: Added.

PR #78, opened 2023-01-12
Import implementation of `ExperimentAxisQuery`.

thetorpedodog:
As discussed in https://github.com/single-cell-data/SOMA/pull/76, the full implementation of `ExperientAxisQuery` should be provided by the base SOMA implementation since it does not vary from storage engine to storage engine. This copies most of the implementation from the existing TileDB version, with some reorganization to make the main query class a little smaller:

- Pulls more of the caching-specific behavior of join IDs into the `_JoinIDCache` class.
- Pulls more indexing-related work into the `_AxisIndexer` class.
  - Makes formerly-public `AxisIndexer` private. (If this should be public this can be easily reverted.)
  - Renames methods to `by_[axis]`.
  - Separates out index caching logic from main methods.
- Uses an `_Axis` enum internally for axis selection.
- Shortens `AxisQueryResult` name.


python-spec/src/somacore/query/query.py
```diff
@@ -110,3 +185,378 @@ def __del__(self) -> None:
         sdel = getattr(super(), "__del__", lambda: None)
         sdel()
         self.close()
+
+    # Internals
+
+    def read(
```
bkmartinjr, line 191: Recommend prefixing this method with an underscore as it is internals.  I originally omitted the underscore, as I wasn't sure if it was internal or public API

I am already confident that I need to significantly restructure `read()` and how it is used by `to_anndata()` for performance reasons.

python-spec/src/somacore/query/query.py
```diff
@@ -110,3 +185,378 @@ def __del__(self) -> None:
...
+        :param X_name: The name of the X layer to read and return
+            in the ``AnnData.X`` slot
+        :param column_names: Specify which column names in ``var`` and ``obs``
+            dataframes to read and return.
+        :param X_layers: Addtional X layers read read and return in the
+            ``AnnData.layers`` slot
+
+        """
+        column_names = column_names or AxisColumnNames(obs=None, var=None)
+        x_collection = self.experiment.ms[self.measurement_name].X
```
bkmartinjr, line 213: why not use `self._ms` property?

python-spec/src/somacore/query/query.py
```diff
@@ -110,3 +185,378 @@ def __del__(self) -> None:
...
+        if self._threadpool_ is None:
+            # TODO: the user should be able to set their own threadpool, a la asyncio's
+            # loop.set_default_executor().  This is important for managing the level of
+            # concurrency, etc.
+            self._threadpool_ = futures.ThreadPoolExecutor()
+        return self._threadpool_
+
+
+@attrs.define(frozen=True)
+class AxisQueryResult:
```
bkmartinjr, line 354: This is (I believe) a private class and should probably have a underscore-prefixed name

thetorpedodog: ibid.

bkmartinjr (commented):
nit: suggest adding `[lifecycle: experimental]` tags to public methods (not just the top-level class)

bkmartinjr (changes requested):
Two changes suggested:
* everything implementing the `to_anndata` code path (eg, `read`, `AxisQueryResult`, etc should have an under-score prefixed name as it is currently just private implementation, and very likely to change for performance reasons.
* suggest pasting the lifecycle maturity tags on the public methods (eg, `lifecycle: experimental`)

thetorpedodog (commented):
> nit: suggest adding `[lifecycle: experimental]` tags to public methods (not just the top-level class)

Added.

bkmartinjr (approved):

PR #79, opened 2023-01-13
Add missing exported classes to somacore

johnkerl:
We need to export `Measurement` and `Experiment` in addition to `Collection`. While it is true the first two are specializations of the third, they are not all interchangeable. Specifically, we use `soma_type: Final = "SOMAMeasurement"` and `soma_type: Final = "SOMAExperiment"` as inherited in their analogous TileDB-SOMA classes. If `TileDB-SOMA`'s `Measurement` inherits from `somacore.Collection` then its `soma_type` will be wrong; which is of course unacceptable.

Context: https://github.com/single-cell-data/TileDB-SOMA/issues/669

### DIFF
--- a/python-spec/src/somacore/__init__.py
+++ b/python-spec/src/somacore/__init__.py
@@ -15,6 +15,8 @@ __version_tuple__ = _version.version_tuple
 
 SOMAObject = base.SOMAObject
 Collection = base.Collection
+Experiment = base.Experiment
+Measurement = base.Measurement
 
 SimpleCollection = ephemeral.SimpleCollection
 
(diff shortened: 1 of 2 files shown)


johnkerl:
@thetorpedodog https://github.com/single-cell-data/TileDB-SOMA/pull/678 is passing format checks and unit tests in my sandbox with this PR in place ... how should we proceed?

johnkerl:
No longer needed on https://github.com/single-cell-data/TileDB-SOMA/pull/678

* I had issues with https://github.com/single-cell-data/TileDB-SOMA/pull/678 and `0.0.0a2` of `somacore`
* Then I went to `main` of `somacore` and made these mods
* But there was a lot of stuff between `0.0.0a2` and this week's `main` of `somacore`, more than the little bit I added between `main` of `somacore` and this PR

With @thetorpedodog 's able help we were able to get https://github.com/single-cell-data/TileDB-SOMA/pull/678 working with `0.0.0a2` of `somacore`.

PR #80, opened 2023-01-14
Move basic Python configuration to the root of the repo.

thetorpedodog:
After way too much trying to figure stuff out I realized: why not just put the `pyproject.toml` right at the root of the repo? This eliminates all the `#subdirectory=...` when doing `pip install`s and also makes `setuptools-scm` work right again. As a bonus, lots of tools now work without any additional configuration and we can just run them as-is.

This moves the bare minimum files needed to the repo root:

- `pyproject.toml`
- `MANIFEST.in`

and updates everything else to make that work right.


thetorpedodog:
haha whoops

thetorpedodog:
we’re back

bkmartinjr (approved):
nice!

PR #81, opened 2023-01-17
Make all coordinates `coords` and add type annotations.

thetorpedodog:
- Changes `ids`, `slices`, and `coords` to all be just `coords`
- Adds appropriate annotations for each type.

---

To see what this will look like when it propagates to `tiledbsoma`: https://github.com/single-cell-data/TileDB-SOMA/commit/ad1cbaeab348ba53e8b10df41098ade574e719ef

### DIFF
--- a/python-spec/src/somacore/data.py
+++ b/python-spec/src/somacore/data.py
@@ -7,7 +7,7 @@ Default values are provided here as a reference for implementors.
 """
 
 import abc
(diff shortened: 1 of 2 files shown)


bkmartinjr (approved):
LGTM

thetorpedodog:
This is part of https://github.com/single-cell-data/TileDB-SOMA/issues/686, which I hadn’t even seen before I made this change.

PR #82, opened 2023-01-18
Re-enable `pip install --editable`; add mypy pre-commit step.

thetorpedodog:
- In the process of moving python installation config to the root on #80, I missed the `setup.py` file. This re-adds it to make ediable installs work again.
- Adds a mypy pre-commit step.


johnkerl (approved):

bkmartinjr (approved):
Thanks!!!

PR #83, opened 2023-01-18
[python] Fast CSR loading for to_anndata

bkmartinjr:
This PR addresses the peformance issues noted in single-cell-data/TileDB-SOMA#719.  In particular, loading large AnnData requires CSR sparse matrices, which are very slow to convert from COO to CSR.  This PR adds a fast-path converter for Python, implementing two changes:
* parallelized COO to CSR conversion
* eager reading of the COO data from SOMA, allowing some concurrent COO-to-CSR work to overlap with the data reading.

In addition, this change uses substantially less memory for the conversion, allowing larger datasets to be loaded into CSR, and therefore into AnnData.

Before/after benchmarks show 1.5-4X speed-ups when the work fits in RAM (tested on r6i instance types with data on S3).  In cases where paging occured, speed-ups were dramatically larger (e.g., 20X).

See also, related PR single-cell-data/TileDB-SOMA#745

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ readme = "./python-spec/README.md"
 dependencies = [
   "anndata",
   "attrs>=22.1",
+  "numba",
   "numpy>=1.21",
   "pandas",
   "pyarrow",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,5 +46,5 @@ single_line_exclusions = ["typing", "typing_extensions"]
 
 [[tool.mypy.overrides]]
 # These dependencies do not currently have canonical type stubs.
-module = ["anndata", "pandas", "pyarrow", "scipy"]
+module = ["anndata", "numba", "pandas", "pyarrow", "scipy"]
 ignore_missing_imports = true
--- a/python-spec/requirements-py3.10.txt
+++ b/python-spec/requirements-py3.10.txt
@@ -2,7 +2,8 @@ anndata==0.8.0
 attrs==22.2.0
 h5py==3.7.0
 natsort==8.2.0
-numpy==1.24.1
+numba==0.56.4
+numpy==1.23.5
 packaging==23.0
 pandas==1.5.2
 pyarrow==10.0.1
--- a/python-spec/src/somacore/query/eager_iter.py
+++ b/python-spec/src/somacore/query/eager_iter.py
@@ -30,9 +30,10 @@ class _EagerIterator(Iterator[_T]):
             self._pool.shutdown()
 
     def __del__(self) -> None:
+        # Ensure the threadpool is cleaned up in the case where the
+        # iterator is not exhausted. For more information on __del__:
+        # https://docs.python.org/3/reference/datamodel.html#object.__del__
+
         super_del = getattr(super(), "__del__", lambda: None)
         super_del()
-
-        # admit the condition where the iterator was abandoned without
-        # completely exhausting the iterator
         self._cleanup()
--- a/python-spec/src/somacore/query/eager_iter.py
+++ b/python-spec/src/somacore/query/eager_iter.py
@@ -4,7 +4,7 @@ from typing import Iterator, Optional, TypeVar
 _T = TypeVar("_T")
 
 
-class EagerIterator(Iterator[_T]):
+class _EagerIterator(Iterator[_T]):
     def __init__(
         self,
         iterator: Iterator[_T],
--- a/python-spec/src/somacore/query/eager_iter.py
+++ b/python-spec/src/somacore/query/eager_iter.py
@@ -1,0 +1,38 @@
+from concurrent import futures
+from typing import Iterator, Optional, TypeVar
+
+_T = TypeVar("_T")
+
+
+class EagerIterator(Iterator[_T]):
+    def __init__(
+        self,
+        iterator: Iterator[_T],
+        pool: Optional[futures.Executor] = None,
+    ):
+        super().__init__()
+        self.iterator = iterator
+        self._pool = pool or futures.ThreadPoolExecutor()
+        self._own_pool = pool is None
+        self.future: futures.Future[_T] = self._pool.submit(next, self.iterator)  # type: ignore
+
+    def __next__(self) -> _T:
+        try:
+            res: _T = self.future.result()
+            self.future = self._pool.submit(next, self.iterator)  # type: ignore
+            return res
+        except StopIteration:
+            self._cleanup()
+            raise
+
+    def _cleanup(self) -> None:
+        if self._own_pool:
+            self._pool.shutdown()
+
+    def __del__(self) -> None:
+        super_del = getattr(super(), "__del__", lambda: None)
+        super_del()
+
+        # admit the condition where the iterator was abandoned without
+        # completely exhausting the iterator
+        self._cleanup()
--- a/python-spec/src/somacore/query/eager_iter.py
+++ b/python-spec/src/somacore/query/eager_iter.py
@@ -14,12 +14,12 @@ class _EagerIterator(Iterator[_T]):
         self.iterator = iterator
         self._pool = pool or futures.ThreadPoolExecutor()
         self._own_pool = pool is None
-        self.future: futures.Future[_T] = self._pool.submit(next, self.iterator)  # type: ignore
+        self._future: futures.Future[_T] = self._pool.submit(next, self.iterator)  # type: ignore
 
     def __next__(self) -> _T:
         try:
-            res: _T = self.future.result()
-            self.future = self._pool.submit(next, self.iterator)  # type: ignore
+            res: _T = self._future.result()
+            self._future = self._pool.submit(next, self.iterator)  # type: ignore
             return res
         except StopIteration:
             self._cleanup()
--- a/python-spec/src/somacore/query/fast_csr.py
+++ b/python-spec/src/somacore/query/fast_csr.py
@@ -1,0 +1,290 @@
+import concurrent.futures as futures
+import os
+from typing import Any, List, Tuple, Type, cast
+
+import numba
+import numpy as np
+import numpy.typing as npt
+import pandas as pd
+import pyarrow as pa
+from scipy import sparse
+
+from somacore.data import SparseNDArray
+from .eager_iter import EagerIterator
+
+
+@numba.jit(nopython=True, nogil=True)  # type: ignore
+def _accum_row_length(row_length, row_ind):
+    for i in range(len(row_ind)):
+        row_length[row_ind[i]] += 1
+    return None
+
+
+@numba.jit(nopython=True, nogil=True)  # type: ignore
+def _copy_chunk_range(
+    data_chunk,
+    row_ind_chunk,
+    col_ind_chunk,
+    data,
+    indices,
+    indptr,
+    row_rng_mask,
+    row_rng_val,
+):
+    for n in range(len(data_chunk)):
+        row = row_ind_chunk[n]
+        if (row & row_rng_mask) != row_rng_val:
+            continue
+        ptr = indptr[row]
+        indices[ptr] = col_ind_chunk[n]
+        data[ptr] = data_chunk[n]
+        indptr[row] += 1
+    return None
+
+
+@numba.jit(nopython=True, nogil=True)  # type: ignore
+def _copy_chunklist_range(
+    chunk_list: numba.typed.List,
+    data,
+    indices,
+    indptr,
+    row_rng_mask_bits,
+    job,
+):
+    row_rng_mask = (2**64 - 1) >> row_rng_mask_bits << row_rng_mask_bits
+    row_rng_val = job << row_rng_mask_bits
+    for data_chunk, row_ind_chunk, col_ind_chunk in chunk_list:
+        _copy_chunk_range(
+            data_chunk,
+            row_ind_chunk,
+            col_ind_chunk,
+            data,
+            indices,
+            indptr,
+            row_rng_mask,
+            row_rng_val,
+        )
+    return None
+
+
+@numba.jit(nopython=True, nogil=True)  # type: ignore
+def _finalize_indptr(indptr):
+    prev = 0
+    for r in range(len(indptr)):
+        t = indptr[r]
+        indptr[r] = prev
+        prev = t
+    return None
+
+
+def _select_dtype(
+    maxval: int,
+) -> Type[np.signedinteger[Any]]:
+    """
+    Ascertain the "best" dtype for a zero-based index. Given our
+    goal of minimizing memory use, "best" is currently defined as
+    smallest.
+    """
+    if maxval > np.iinfo(np.int32).max:
+        return np.int64
+    else:
+        return np.int32
+
+
+def _reindex_and_cast(
+    index: pd.Index, ids: npt.NDArray[np.int64], target_dtype: npt.DTypeLike
+) -> npt.NDArray[np.int64]:
+    return cast(
+        npt.NDArray[np.int64], index.get_indexer(ids).astype(target_dtype, copy=False)
+    )
+
+
+class CSRAccumulator:
+    """
+    Fast accumulator of a CSR, based upon COO input.
+    """
+
+    def __init__(
+        self,
+        obs_joinids: npt.NDArray[np.int64],
+        var_joinids: npt.NDArray[np.int64],
+        pool: futures.Executor,
+    ):
+        self.obs_joinids = obs_joinids
+        self.var_joinids = var_joinids
+        self.pool = pool
+
+        self.shape: Tuple[int, int] = (len(self.obs_joinids), len(self.var_joinids))
+        self.obs_indexer: pd.Index = pd.Index(self.obs_joinids)
+        self.var_indexer: pd.Index = pd.Index(self.var_joinids)
+        self.row_length: npt.NDArray[np.int64] = np.zeros(
+            (self.shape[0],), dtype=_select_dtype(self.shape[1])
+        )
+
+        # COO accumulated chunks, stored as list of triples (data, row_ind, col_ind)
+        self.coo_chunks: List[
+            Tuple[
+                npt.NDArray[np.number[Any]],
+                npt.NDArray[np.integer[Any]],
+                npt.NDArray[np.integer[Any]],
+            ]
+        ] = []
+
+    def append(
+        self, data: pa.Array, row_joinids: pa.Array, col_joinids: pa.Array
+    ) -> None:
+        """
+        At accumulation time, do several things:
+        * re-index to positional indices, and if possible, cast to smaller dtype
+          to minimize memory footprint (at cost of some amount of time)
+        * accumulate column counts by row, i.e., build the basis of the indptr
+        * cache the tuple of data, row, col
+        """
+        rows_future = self.pool.submit(
+            _reindex_and_cast,
+            self.obs_indexer,
+            row_joinids.to_numpy(),
+            _select_dtype(self.shape[0]),
+        )
+        cols_future = self.pool.submit(
+            _reindex_and_cast,
+            self.var_indexer,
+            col_joinids.to_numpy(),
+            _select_dtype(self.shape[1]),
+        )
+        row_ind = rows_future.result()
+        col_ind = cols_future.result()
+        self.coo_chunks.append((data.to_numpy(), row_ind, col_ind))
+        _accum_row_length(self.row_length, row_ind)
+
+    def finalize(
+        self,
+    ) -> Tuple[
+        npt.NDArray[np.number[Any]],  # data
+        npt.NDArray[np.integer[Any]],  # indptr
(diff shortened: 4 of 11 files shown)


python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,291 @@
...
+
+def _reindex_and_cast(
+    index: pd.Index, ids: npt.NDArray[np.int64], target_dtype: npt.DTypeLike
+) -> npt.NDArray[np.int64]:
+    return cast(
+        npt.NDArray[np.int64], index.get_indexer(ids).astype(target_dtype, copy=False)
+    )
+
+
+class CSRAccumulator:
```
thetorpedodog, line 103: Since this class doesn’t need the presence of all the above private functions until its own methods are called, I would put this first in the file so the most important stuff is above the fold (and can give useful context to the internal helpers defined thereafter).

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,291 @@
...
+            _select_dtype(self.shape[1]),
+        )
+        row_ind = rows_future.result()
+        col_ind = cols_future.result()
+        self.coo_chunks.append((data.to_numpy(), row_ind, col_ind))
+        _accum_row_length(self.row_length, row_ind)
+
+    def finalize(
+        self,
+    ) -> Tuple[
```
thetorpedodog, line 163: Would it make sense to return this as an attrs class or namedtuple so that these fields can have a name rather than / in addition to being positional?

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,291 @@
...
+                self.pool.submit(
+                    _copy_chunklist_range,
+                    chunk_list,
+                    data,
+                    indices,
+                    indptr,
+                    row_rng_mask_bits,
+                    job,
+                )
+                for job in range(0, n_jobs)
```
thetorpedodog, line 199:
```suggestion
                for job in range(n_jobs)
```

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    matrix = sparse.csr_matrix.__new__(sparse.csr_matrix)
+    matrix.data = data
+    matrix.indices = indices
+    matrix.indptr = indptr
+    matrix._shape = shape
+    return matrix
+
+
+def fast_SparseCSRMatrix_to_scipy(csr: pa.SparseCSRMatrix) -> sparse.csr_matrix:
```
thetorpedodog, line 282: `(arrow_?)csr_to_scipy`? (we already know it’s fast from the module name)

thetorpedodog (commented):
I have some style suggestions and related concerns here, but the overall structure of the code looks pretty good. No major changes, just some polishing.

python-spec/src/somacore/query/eager_iter.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(
+        self,
+        iterator: Iterator[_T],
+        pool: Optional[futures.Executor] = None,
+    ):
+        super().__init__()
+        self.iterator = iterator
+        self._pool = pool or futures.ThreadPoolExecutor()
+        self._own_pool = pool is None
+        self.future: futures.Future[_T] = self._pool.submit(next, self.iterator)  # type: ignore
```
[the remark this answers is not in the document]
bkmartinjr: By privateish - you mean preface the name with an underscore?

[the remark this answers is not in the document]
thetorpedodog: yep

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,291 @@
...
+from typing import List, Tuple, Type, cast
+
+import numba
+import numpy as np
+import numpy.typing as npt
+import pandas as pd
+import pyarrow as pa
+from scipy import sparse
+
+from somacore.data import SparseNDArray
```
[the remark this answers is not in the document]
bkmartinjr: the name `data` collided with much of my code, which uses it as a variable (becaues SciPy and Arrow use it).  I was trying to not pollute the name space :-)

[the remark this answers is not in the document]
bkmartinjr: Also, the change from a dot-import requires that I export the private _EagerIterator in the query/__init__.py, which I would rather not do (as it is private)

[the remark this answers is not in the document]
bkmartinjr: So...LMK how strongly you feel about these two, as they both require changes I consider sub-optimal.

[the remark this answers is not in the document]
thetorpedodog: `import somacore.data as scd`, maybe?

There should be no export changes required for the `eager_iter` import; `from somacore.query import eager_iter` works as-is.

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,291 @@
...
+
+from .eager_iter import EagerIterator
+
+
+@numba.jit(nopython=True, nogil=True)  # type: ignore
+def _accum_row_length(row_length, row_ind):
+    for i in range(len(row_ind)):
+        row_length[row_ind[i]] += 1
+    return None
```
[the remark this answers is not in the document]
bkmartinjr: No.  I just find them clarifying.

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: the overhead didn't seem worth it given that it is a private interface.  But sure, why not ... :-)

python-spec/src/somacore/query/fast_csr.py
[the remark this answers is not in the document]
bkmartinjr: ty - I copied your code!

bkmartinjr:
@thetorpedodog - OK, I think I hit all of your (excellent) feedback.

python-spec/src/somacore/query/eager_iter.py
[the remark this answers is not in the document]
thetorpedodog: Did you miss this one?

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,300 @@
...
+import numpy.typing as npt
+import pandas as pd
+import pyarrow as pa
+from scipy import sparse
+
+import somacore.data as scd
+from somacore.query import eager_iter
+
+
+class CSRAccumulatorFinalResult(NamedTuple):
```
thetorpedodog, line 16: If this is only internal, maybe prefix it with _?

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,300 @@
...
+        self.shape: Tuple[int, int] = (len(self.obs_joinids), len(self.var_joinids))
+        self.obs_indexer: pd.Index = pd.Index(self.obs_joinids)
+        self.var_indexer: pd.Index = pd.Index(self.var_joinids)
+        self.row_length: npt.NDArray[np.int64] = np.zeros(
+            (self.shape[0],), dtype=_select_dtype(self.shape[1])
+        )
+
+        # COO accumulated chunks, stored as list of triples (data, row_ind, col_ind)
+        self.coo_chunks: List[
+            Tuple[
```
thetorpedodog, line 52: potential thought: make this a CSRAccumulatorFinalResult, and just don't use the `shape` field?

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,300 @@
...
+    index: pd.Index, ids: npt.NDArray[np.int64], target_dtype: npt.DTypeLike
+) -> npt.NDArray[np.int64]:
+    return cast(
+        npt.NDArray[np.int64], index.get_indexer(ids).astype(target_dtype, copy=False)
+    )
+
+
+def _read_csr(
+    matrix: scd.SparseNDArray, obs_joinids: pa.Array, var_joinids: pa.Array
+) -> Tuple[
```
thetorpedodog, line 217: ibid. re using CSRAccumulatorFinalResult. A more general name might make sense for it. `_DataWithIndices`?

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def _create_scipy_csr_matrix(
+    data: npt.NDArray[np.number],
+    indices: npt.NDArray[np.integer],
+    indptr: npt.NDArray[np.integer],
+    shape: Tuple[int, int],
+) -> sparse.csr_matrix:
+    """
+    Create a Scipy sparse.csr_matrix from
```
thetorpedodog, line 273: unfinished sentence, a crime that I am also often guilty of

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,300 @@
...
+    """
+    Given a 2D SparseNDArray and joinids for the two dimensions, read the
+    slice and return it as an SciPy sparse.csr_matrix.
+    """
+    data, indptr, indices, shape = _read_csr(matrix, obs_joinids, var_joinids)
+    csr = pa.SparseCSRMatrix.from_numpy(data, indptr, indices, shape=shape)
+    return csr
+
+
+def _create_scipy_csr_matrix(
```
thetorpedodog, line 266: make this a method of CAFR, maybe?

python-spec/src/somacore/query/query.py
```diff
@@ -156,14 +158,20 @@ def to_anndata(
...
-        query_result = self._read(
+        # query_result = self._read(
+        #     X_name,
+        #     column_names=column_names or AxisColumnNames(obs=None, var=None),
+        #     X_layers=X_layers,
+        # )
+
+        # # AnnData uses positional indexing
+        # return self._indexer.rewrite(query_result).to_anndata()
+
```
thetorpedodog, line 169: probably don't need this any longer.

thetorpedodog (approved):
a few minor, non-essential suggestions, which you are free to take or leave. looks good!

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: I used a namedtuple.  LMK if this was along the lines of what you were thinking.

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: they are quite different.  One is COO, the other is CSR.  I would find this change very confusing.

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: Same issue - they are completely different matrix formats.  COO vs CSR

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: it has general (internal) utility - I imagine soma.io is going to want to pick up some of this stuff, as it has exactly the same issue.

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: If you are asking for this to be put into a named tuple, I can add that - but it will need to be a different type than CSRAccumulatorFinalResult. I may also need to convert to an actual tuple for numba, as I believe numba.List only supports native tuples.

python-spec/src/somacore/query/eager_iter.py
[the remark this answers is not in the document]
bkmartinjr: ah, I misunderstood the original comment!  Fixing.

thetorpedodog:
Still looks good.

pyproject.toml
```diff
@@ -13,6 +13,7 @@ readme = "./python-spec/README.md"
 dependencies = [
   "anndata",
   "attrs>=22.1",
+  "numba",
```
johnkerl, line 16: `numba` doesn't work with Python 3.11 -- but, we're alreaady getting `numba` as an _indirect_ dependency so this is nothing new. (See also https://github.com/single-cell-data/TileDB-SOMA/issues/544)

python-spec/requirements-py3.10.txt
```diff
@@ -2,7 +2,8 @@ anndata==0.8.0
 attrs==22.2.0
 h5py==3.7.0
 natsort==8.2.0
-numpy==1.24.1
+numba==0.56.4
```
johnkerl, line 5: I'm curious why `==` down to the micro -- my general experience is this is "overpinning" but clearly not the case here

Not specific to this PR; I see other deps are that way too ...

Anyway, just seeking to learn -- cc @thetorpedodog

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,299 @@
...
+    matrix.data = data
+    matrix.indices = indices
+    matrix.indptr = indptr
+    matrix._shape = shape
+    return matrix
+
+
+def _SparseCSRMatrix_to_scipy(csr: pa.SparseCSRMatrix) -> sparse.csr_matrix:
+    """
+    Convert Arrow SparseCSRMatrix to scipy.sparse.csr_matrix. Semantically the same
```
johnkerl, line 292: Convert pyarrow.SparseCSRMatrix to scipy.sparse.csr_matrix ...

python-spec/src/somacore/query/eager_iter.py
```diff
@@ -0,0 +1,38 @@
...
+            return res
+        except StopIteration:
+            self._cleanup()
+            raise
+
+    def _cleanup(self) -> None:
+        if self._own_pool:
+            self._pool.shutdown()
+
+    def __del__(self) -> None:
```
johnkerl, line 32: maybe add a helpful comment to the reader why a `__del__` operator is even on the table here -- the reader (including myself) would surely learn something, at just the cost of a sentence or two here

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,299 @@
...
+        self.pool = pool
+
+        self.shape: Tuple[int, int] = (len(self.obs_joinids), len(self.var_joinids))
+        self.obs_indexer: pd.Index = pd.Index(self.obs_joinids)
+        self.var_indexer: pd.Index = pd.Index(self.var_joinids)
+        self.row_length: npt.NDArray[np.int64] = np.zeros(
+            (self.shape[0],), dtype=_select_dtype(self.shape[1])
+        )
+
+        # COO accumulated chunks, stored as list of triples (data, row_ind, col_ind)
```
johnkerl, line 50: IJV (rather than VIJ) is pretty standard -- any particular reason to be different here?

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,299 @@
...
+            pool=pool,
+        ):
+            acc.append(tbl["soma_data"], tbl["soma_dim_0"], tbl["soma_dim_1"])
+
+        data, indptr, indices, shape = acc.finalize()
+
+    return data, indptr, indices, shape
+
+
+def read_scipy_csr(
```
johnkerl, line 242: It would do a lot for readability to put the main entrypoint methods (`read_scipy_csr` and `read_arrow_csr`) at the top of the file -- which is where people generally begin reading.

python-spec/src/somacore/query/fast_csr.py
johnkerl: @bkmartinjr 100%

I have an open to-do for outgest (some `main-old` feedback) & my first thought was using more `ThreadPoolExecutor` there. But maybe this is exactly what it really needs.

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,299 @@
...
+    csr = _create_scipy_csr_matrix(data, indices, indptr, shape=shape)
+    return csr
+
+
+def read_arrow_csr(
+    matrix: scd.SparseNDArray, obs_joinids: pa.Array, var_joinids: pa.Array
+) -> pa.SparseCSRMatrix:
+    """
+    Given a 2D SparseNDArray and joinids for the two dimensions, read the
+    slice and return it as an SciPy sparse.csr_matrix.
```
johnkerl, line 259: as a `scipy.sparse.csr_matrix`

... except I think you mean `pyarrow.SparseCSRMatrix`

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,299 @@
...
+def _read_csr(
+    matrix: scd.SparseNDArray, obs_joinids: pa.Array, var_joinids: pa.Array
+) -> Tuple[
+    npt.NDArray[np.number],  # data
+    npt.NDArray[np.integer],  # indptr
+    npt.NDArray[np.integer],  # indices
+    Tuple[int, int],  # shape
+]:
+    if not isinstance(matrix, scd.SparseNDArray) or matrix.ndim != 2:
+        raise TypeError("read_scipy_csr can only read from a 2D SparseNDArray")
```
johnkerl, line 224: `_read_csr can only ...`

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,299 @@
...
+@numba.jit(nopython=True, nogil=True)  # type: ignore
+def _copy_chunklist_range(
+    chunk_list: numba.typed.List,
+    data: npt.NDArray[np.number],
+    indices: npt.NDArray[np.signedinteger],
+    indptr: npt.NDArray[np.signedinteger],
+    row_rng_mask_bits: int,
+    job: int,
+):
+    row_rng_mask = (2**64 - 1) >> row_rng_mask_bits << row_rng_mask_bits
```
johnkerl, line 169: do we need to assert `row_rng_mask_bits` is less than 64? here it's 18 (currently) but if any well-meaning soul every changes that value down there without looking up here, i suspect they'll be doing a bit of debugging to figure out what's going wrong ...

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,299 @@
...
+        # cumsum row lengths to get indptr
+        indptr = np.empty((self.shape[0] + 1,), dtype=index_dtype)
+        indptr[0:1] = 0
+        np.cumsum(self.row_length, out=indptr[1:])
+
+        # Parallel copy of data and column indices
+        indices = np.empty((nnz,), dtype=index_dtype)
+        data = np.empty((nnz,), dtype=self.coo_chunks[0][0].dtype)
+
+        row_rng_mask_bits = 18
```
johnkerl, line 108: why is this a good number? a sentence or two of context would be great.

johnkerl (changes requested):

python-spec/requirements-py3.10.txt
bkmartinjr: You will have to ask @thetorpedodog 

My guess is that he wants to control the deps for testing purposes.  But either way, I did not introduce this pattern with this PR.

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: pyarrow is simply a wrapper on Arrow.

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: no reason.  they are local variables, so I didn't think it was unobvious.  Will reorder.

python-spec/requirements-py3.10.txt
johnkerl: > My question here is not specific to this PR; I see other deps are that way too ...
> 
> Anyway, I'm just seeking to learn

> I did not introduce this pattern with this PR.

We are in 100% agreement here

python-spec/src/somacore/query/fast_csr.py
```diff
@@ -0,0 +1,310 @@
...
+        # cumsum row lengths to get indptr
+        indptr = np.empty((self.shape[0] + 1,), dtype=index_dtype)
+        indptr[0:1] = 0
+        np.cumsum(self.row_length, out=indptr[1:])
+
+        # Parallel copy of data and column indices
+        indices = np.empty((nnz,), dtype=index_dtype)
+        data = np.empty((nnz,), dtype=self.coo_chunks[0][2].dtype)
+
+        # empirically determined value. Needs to be large enough for reasonable
```
johnkerl, line 137: Super-nit: capitalize Empirically at start of sentence

johnkerl (approved):
🚢 🚢 🚢

python-spec/src/somacore/query/fast_csr.py
bkmartinjr: super nit, but super easy - fixed!

bkmartinjr:
Related issue with scipy.sparse performance:  scipy/scipy#11496

Issue #84, opened 2023-01-18
Switch from `ndims` to `ndim`


johnkerl:
From https://github.com/single-cell-data/TileDB-SOMA/issues/744, as it turns out this is spec-only.

Context there: NumPy, SciPy, Pandas, et al. use `ndim` -- so should we. Affects spec, Python impl, and R impl.

This wasn't caught sooner since, unlike other properties in `NDArray`, `ndims` does not have the `@abc.abstractmethod`:

https://github.com/single-cell-data/SOMA/blob/python-v0.0.0a3/python-spec/src/somacore/data.py#L84-L99

PR #85, opened 2023-01-18
Change NDArray ndims to ndim

johnkerl:
Addresses #84

### DIFF
--- a/python-spec/src/somacore/data.py
+++ b/python-spec/src/somacore/data.py
@@ -88,10 +88,9 @@ class NDArray(base.SOMAObject, metaclass=abc.ABCMeta):
         raise NotImplementedError()
 
     @property
-    @abc.abstractmethod
     def ndim(self) -> int:
         """The number of dimensions in this array."""
-        raise NotImplementedError()
+        return len(self.shape)
 
     @property
     @abc.abstractmethod
--- a/python-spec/src/somacore/data.py
+++ b/python-spec/src/somacore/data.py
@@ -88,8 +88,9 @@ class NDArray(base.SOMAObject, metaclass=abc.ABCMeta):
         raise NotImplementedError()
 
     @property
-    def ndims(self) -> int:
-        """The Number of Dimensions in this array."""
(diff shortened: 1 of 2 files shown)


python-spec/src/somacore/data.py
```diff
@@ -88,8 +88,9 @@ def shape(self) -> Tuple[int, ...]:
         raise NotImplementedError()
 
     @property
-    def ndims(self) -> int:
-        """The Number of Dimensions in this array."""
+    @abc.abstractmethod
```
thetorpedodog, line 91: Since this is the implementation of ndim, this should not be marked as abstract.

thetorpedodog (approved):

bkmartinjr (approved):
LGTM. 

May also want to update `abstract_specfication.md`

python-spec/src/somacore/data.py
```diff
@@ -91,7 +91,7 @@ def shape(self) -> Tuple[int, ...]:
     @abc.abstractmethod
     def ndim(self) -> int:
         """The number of dimensions in this array."""
-        return len(self.shape)
+        raise NotImplementedError()
```
thetorpedodog, line 94: This should still be `len(self.shape)`; there’s no need to make downstream users implement this.

thetorpedodog (changes requested):

python-spec/src/somacore/data.py
thetorpedodog: Did not see the earlier comment.

thetorpedodog: Actually it looks like this still can be `len(self.shape)`. An implementation *can* override this, but it should always be valid.

thetorpedodog (approved):

python-spec/src/somacore/data.py
thetorpedodog: What would need changing there? It returns a tuple whose length will be `ndim`, so taking its `len` should be fine (as the spec implies). Am I missing something?

PR #88, opened 2023-01-19
Create `add` method for base Collection type

thetorpedodog:
Also pulls the `set` method for Collections from TileDB-SOMA, allowing a user to decide whether a collection entry should be stored with an absolute or relative URI.

This is the first step in reworking the SOMA object creation and opening protocols with the eventual goal that (unless the user does something weird) user code should only ever have to deal with fully-set-up SOMA objects and collections.

---

My biggest question here is: Does it make sense to import the `set` method from tiledbsoma?

### DIFF
--- a/python-spec/src/somacore/_wrap.py
+++ b/python-spec/src/somacore/_wrap.py
@@ -18,6 +18,7 @@ import attrs
 from typing_extensions import Final
 
 from somacore import base
+from somacore import options
 
 _ST = TypeVar("_ST", bound=base.SOMAObject)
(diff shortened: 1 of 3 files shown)


bkmartinjr (approved):
This seems good.

To your question:
> My biggest question here is: Does it make sense to import the set method from tiledbsoma?

I think it does, as I believe the issue of relative URI is going to be common and should be in the base API.  If a platform/implementation does not use it, it can easily ignore the param while implementing its own semantics.

johnkerl (approved):

PR #89, opened 2023-01-19
fix issue TileDB-SOMA#755

bkmartinjr:
Context: single-cell-data/TileDB-SOMA#755 and chanzuckerberg/cell-census#56

This PR modifies the `ExperimentAxisQuery.to_anndata()` to return columns consistent with other API, i.e., returned obs/var columns will be the union of columns explicitly requested via `column_names` parameter, *and* columns referenced in the AxisQuery `value_filter`.

Fixes single-cell-data/TileDB-SOMA#755

### DIFF
--- a/python-spec/src/somacore/query/query.py
+++ b/python-spec/src/somacore/query/query.py
@@ -280,6 +280,7 @@ class ExperimentAxisQuery:
         # If we can cache join IDs, prepare to add them to the cache.
         joinids_cached = self._joinids._is_cached(axis)
         query_columns = column_names
+        added_soma_joinid_to_columns = False
         if (
             not joinids_cached
             and column_names is not None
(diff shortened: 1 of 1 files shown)


python-spec/src/somacore/query/query.py
```diff
@@ -307,10 +309,10 @@ def _read_axis_dataframe(
                 arrow_table.column("soma_joinid").combine_chunks(),
             )
 
-        # Ensure that we return the exact columns the caller was expecting,
-        # even if we added our own above.
-        if column_names is not None:
-            arrow_table = arrow_table.select(column_names)
+        # Drop soma_joinid column if we added it soley for use in filling
```
thetorpedodog, line 312:
```suggestion
        # Drop soma_joinid column if we added it solely for use in filling
```

thetorpedodog (approved):

johnkerl (approved):

python-spec/src/somacore/query/query.py
bkmartinjr: oops!  fixed

PR #91, opened 2023-01-20
Add ruff (a fast Python linter) to pre-commit.

thetorpedodog:
Shamelessly stolen from the TileDB SOMA repository [1], this change adds ruff to the pre-commit workflow.

Also does some minor changes to placate ruff so we start off with a completely clean slate.

[1] https://github.com/single-cell-data/TileDB-SOMA/pull/751

(cc @gsakkis)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,3 +28,8 @@ repos:
     rev: '5.11.4'
     hooks:
       - id: isort
+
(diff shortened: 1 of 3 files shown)


johnkerl (approved):

gsakkis:
Btw @ihnorton [pointed out](https://github.com/TileDB-Inc/TileDB-Py/pull/1597#issuecomment-1400300846) that `ruff` can subsume `isort`, you may remove the latter if you want.

PR #93, opened 2023-01-20
Add `uri` property to base SOMA objects.

thetorpedodog:


python-spec/testing/test_wrap.py
```diff
@@ -9,6 +9,9 @@ def test_sanity(self):
         backing = ephemeral.SimpleCollection()
         wrapped = _wrap.CollectionProxy(backing)
 
+        back_repr = repr(backing)
```
johnkerl, line 12: nit: `back_repr` -> `backing_repr`

johnkerl (approved):
🐑

python-spec/testing/test_wrap.py
thetorpedodog: done.

PR #94, opened 2023-01-23
Split apart experiment and measurement to break import cycle

thetorpedodog:
To eliminate a troublesome import cycle between query and composed, this breaks those two into separate parts.  query now defines the subset of Experiment that it uses in the form of a Protocol, to avoid depending directly on experiment.py.


python-spec/src/somacore/experiment.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class Experiment(_wrap.CollectionProxy):
+    """A set of observations defined by a DataFrame, with measurements."""
+
+    obs = _wrap.item(data.DataFrame)
+    """Primary observations on the observation axis.
+
+    The contents of the ``soma_joinid`` pseudo-column define the observation
+    index domain, i.e. ``obsid``. All observations for the experiment must be
```
johnkerl, line 21: `obs_id` is more standard

johnkerl (approved):

thetorpedodog:
@johnkerl Please take another look; I added one more minor reorg here.

Neither of these changes affect the public API or the namespace available in `somacore`.

bkmartinjr (approved):
LGTM

PR #95, opened 2023-01-23
Rework Experiment and Measurement to be mixins.

thetorpedodog:
Previously Experiment and Measurement were implemented as wrappers around SOMA objects. However, as the complexity of the SOMA interfaces increases (and we want to ensure that they are extensible by implementations), it is clear that they make more sense as mixins.

This follows the inspiration of `pathlib`, where extra behaviors are specified in a mixin, and a specialization inherits both the implementation class (i.e., a concrete `Collection`) and the extra behaviors (e.g. `Experiment` or `Measurement`).

The parent of the `Experiment` and `Measurement` mixins are specified as `MutableMapping`s instead of `Collection`s to avoid an inheritance conflict in the Python interpreter:

`TypeError: multiple bases have instance lay-out conflict`.


python-spec/testing/test_collection.py
```diff
@@ -19,3 +21,12 @@ def test_basic(self):
         md["hello"] = "world"
 
         self.assertEqual("world", coll.metadata["hello"])
+
+    def test_mro(self):
```
johnkerl, line 25: please spell out (as the function name) the abbreviation `mro`

johnkerl (changes requested):

python-spec/testing/test_collection.py
thetorpedodog: Since [the term `mro` is used in the Python object model](https://docs.python.org/3/library/stdtypes.html#class.__mro__), I think it makes sense to use here while providing its expansion in the comment.

python-spec/src/somacore/_mixin.py
```diff
@@ -0,0 +1,97 @@
...
+_ST = TypeVar("_ST", bound=base.SOMAObject)
+_Coll = MutableMapping[str, _ST]
+_T = TypeVar("_T")
+
+
+@attrs.define()
+class item(Generic[_T]):
+    """Descriptor to transform property access into indexing.
+
+    This descriptor works with :class:`CollectionProxy` to allow for simple
```
bkmartinjr, line 18: this comment seems slightly stale?

bkmartinjr (approved):
This generally looks good to me, with the caveat that I am not expert in this pattern.

python-spec/src/somacore/_mixin.py
thetorpedodog: Oops! You're right.

python-spec/testing/test_collection.py
johnkerl: Please spell out `test_method_resolution_order`

Ours is a package for users with varying strengths, skilled in different languages

While Python-expert software developers may have memorized certain insider acronyms, let's go ahead and take a few keystrokese to communicate clearly to everyone else.

johnkerl (approved):
Async approve as long as my ask is addressed 🙏

python-spec/testing/test_collection.py
thetorpedodog: Given that this is test code, I'm not hugely attached to either; I stuck the MRO acronym into the comment so it is still visible.

PR #96, opened 2023-01-23
Rework object-creation and collection-building protocols

thetorpedodog:
This is (the specification side of) the plan for reworking the way we do creating and building complex SOMA objects. Currently, the primary model in TileDB-SOMA is to create the "leaf" objects, then work your way out to creating the "root" objects. This change is intended to better support a top-down creation model, where a user can create (for example) an Experiment, then create the members inside of it:

```python
new_exp = soma_impl.Experiment.create("uri/of/new/experiment", ...)
# new_exp is opened in write mode

new_exp.add_dataframe("obs", ...)
# an `obs` dataframe is created inside the experiment

# The user can continue to build down into the experiment,
# or any other collection they have opened in write mode.
```

It also includes a changed flow for opening SOMA objects, where the SOMA object should always refer to a known object, rather than potentially being a free-floating URI:

```python
my_coll = soma_impl.Collection.open("/path/to/whatever", ...)
# my_coll is an active, opened collection in read mode.
```

This is also intended to enable a base "smart open" function in the SOMA implementation:

```python
some_obj = soma_impl.open("/path/to/any/soma_object", ...)
# the type of some_obj is identified and an instance of the appropriate
# concrete class is returned.
```

### DIFF
--- a/python-spec/src/somacore/collection.py
+++ b/python-spec/src/somacore/collection.py
@@ -214,6 +214,11 @@ class SimpleCollection(Collection[_ST]):
         return self._metadata
 
     @classmethod
+    def open(cls, *args, **kwargs) -> NoReturn:
+        del args, kwargs  # All unused
+        raise TypeError("SimpleCollections are in-memory only and cannot be opened.")
+
+    @classmethod
     def create(cls, *args, **kwargs) -> "SimpleCollection":
         del args, kwargs  # All unused
         # SimpleCollection is in-memory only, so just return a new empty one.
--- a/python-spec/src/somacore/collection.py
+++ b/python-spec/src/somacore/collection.py
@@ -1,13 +1,27 @@
 import abc
-from typing import Any, Dict, Iterator, MutableMapping, Optional, Type, TypeVar
+from typing import (
+    Any,
+    Dict,
+    Iterator,
+    MutableMapping,
+    NoReturn,
+    Optional,
+    Sequence,
+    Type,
+    TypeVar,
+)
 
+import pyarrow as pa
 from typing_extensions import LiteralString
 
 from . import base
+from . import data
 from . import options
 
 _ST = TypeVar("_ST", bound=base.SOMAObject)
 """Generic type variable for any SOMA object."""
+_CT = TypeVar("_CT", bound="Collection")
+"""Any implementation of a Collection."""
 
 
 class Collection(base.SOMAObject, MutableMapping[str, _ST], metaclass=abc.ABCMeta):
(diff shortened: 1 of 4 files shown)


python-spec/src/somacore/collection.py
```diff
@@ -20,34 +34,47 @@ class Collection(base.SOMAObject, MutableMapping[str, _ST], metaclass=abc.ABCMet
...
-            data = the_collection.create(
-                "data",
-                somacore.DataFrame,
-                uri="file:///custom/path/to/df",
-                platform_config={"somaimpl": ...},
-            )
+            # This will create a basic Collection as a child.
+            sub_child = density.add_collection("sub_child")
 
         By default (in situations where it is possible to do so), the default
```
bkmartinjr, line 79: is this a suggestion/best practice, or a requirement?  Some keys are not legal URI path components (eg, may contain characters illegal in a URI).  If just a best practice, perhaps label it as such.

python-spec/src/somacore/collection.py
```diff
@@ -135,22 +213,30 @@ def uri(self) -> str:
...
+
+    @classmethod
+    def create(cls, *args, **kwargs) -> "SimpleCollection":
+        del args, kwargs  # All unused
+        # SimpleCollection is in-memory only, so just return a new empty one.
+        return cls()
+
+    def add_collection(self, *args, **kwargs) -> NoReturn:
+        del args, kwargs  # All unused
         # TODO: Should we be willing to create Collection-based child elements,
```
bkmartinjr, line 229: I'm struggling to answer this question, as I'm unsure what the use case is for SimpleCollection.  So I think, for now, your simplest-possible-implementation is a good one.

python-spec/src/somacore/collection.py
```diff
@@ -20,34 +34,47 @@ class Collection(base.SOMAObject, MutableMapping[str, _ST], metaclass=abc.ABCMet
...
+        context: Optional[Any] = None,
+    ) -> _CT:
+        """Creates a new Collection at the given URI and returns it.
+
+        The collection will be returned in the opened state.
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def add_collection(
```
bkmartinjr, line 53: the names `create()` vs `add_foo()` seem slightly in opposition.  In both cases, a new persistent object is created, but the term `add` doesn't really convey that (to me).

Assuming our term for construction is `create`, I only have bad naming suggestions, but here goes:
* create_child_collection()
* create_add_collection()

I don't think we should hold up the PR for this - unless somebody has a good naming idea. But `add` didn't really say "i am a factory" method to me.

bkmartinjr (approved):
Very minor comments - nothing that would hold up landing this and continuing to refine.

python-spec/src/somacore/collection.py
johnkerl: @bkmartinjr @thetorpedodog `add_collection` -> `make_collection` maybe -- ?

johnkerl (approved):

python-spec/src/somacore/collection.py
thetorpedodog: I was definitely feeling this as I wrote this but did not have a better idea at the time. Coming back with fresh eyes, I think `add_new_whatever` is a pretty good way to go.

python-spec/src/somacore/collection.py
thetorpedodog: I've added more details about what should be done here:

- Use the key as the sub-URI where possible
- Sanitize it if need be

PR #97, opened 2023-01-24
Make `coords` for reading DataFrame and SparseNDArray optional.

thetorpedodog:
The TileDB implementation currently has `coords` as an optional, default-None value for reads of DataFrames and SparseNDArrays. This updates the ABCs to match that reality.


johnkerl (approved):
@thetorpedodog good timing! Doing gap-analysis work for #643 I noticed that having to say `coords=(slice(None),)` is unwieldy, and regressive with comparison to `main-old`

This will help UX & uptake I believe! :)

bkmartinjr:
Suggest that the same change be made to ExperimentAxisQuery, as we want these to be aligned.

Issue #98, opened 2023-01-24
[python] Also support `coords=None` in `ExperimentAxisQuery`


johnkerl:
_Originally posted by @bkmartinjr in https://github.com/single-cell-data/SOMA/issues/97#issuecomment-1402435287_

thetorpedodog:
Where are we expecting `coords=None` to get passed? As an explicit parameter to `AxisQuery`?

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a6/apis/python/src/tiledbsoma/experiment.py#L12

https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a6/apis/python/src/tiledbsoma/experiment.py#L75-L76

This could now be

```
        obs_query: Union[None, somacore.AxisQuery] = None,
        var_query: Union[None, somacore.AxisQuery] = None,
```

and this
https://github.com/single-cell-data/SOMA/blob/main/python-spec/src/somacore/query/axis.py#L74
could default to `None` rather than `(slice(None),)`

bkmartinjr:
Sorry, I was vague - John translated (ty!!)

I think the `axis_query()` equivalent to the coords default is to make `obs_query` and `var_query` optional, and have that default to `AxisQuery()`.  The original code worked this way, but it disappeared in your refactor.  It is a good UX feature and I had a few complaints about the change already :-)

atolopko-czi:
Agree with Bruce's response.  For the calling code, not specifying a `{obs,var}_query` keyword arg should be the same as including it with a `None` value (`obs_var=None`).  This change broke code we had in Cell Census; we had a method that wanted to be able to just pass through a `None` param value, without having to translate it to `AxisQuery()`.

johnkerl:
I think this is fine now -- whether we spell `coords=None` or not the key ask is for a good "all" default and we have this:

```
>>> exp = soma.Experiment.open("/Users/johnkerl/s/v/pbmc-small-x-csr")

>>> qry.to_anndata(X_name="data")
AnnData object with n_obs × n_vars = 80 × 20
    obs: 'soma_joinid', 'obs_id', 'orig.ident', 'nCount_RNA', 'nFeature_RNA', 'RNA_snn_res.0.8', 'letter.idents', 'groups', 'RNA_snn_res.1'
[...]

PR #99, opened 2023-01-24
Allow passing `None` to `Experiment.axis_query`.

thetorpedodog:
cf. #98

### DIFF
--- a/python-spec/src/somacore/experiment.py
+++ b/python-spec/src/somacore/experiment.py
@@ -1,4 +1,4 @@
-from typing import MutableMapping, TypeVar
+from typing import MutableMapping, Optional, TypeVar
 
 from typing_extensions import Final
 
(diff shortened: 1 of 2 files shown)


python-spec/src/somacore/experiment.py
```diff
@@ -47,8 +47,8 @@ def axis_query(
         self: _Self,
         measurement_name: str,
         *,
-        obs_query: query.AxisQuery = query.AxisQuery(),
-        var_query: query.AxisQuery = query.AxisQuery(),
+        obs_query: Optional[query.AxisQuery] = None,
+        var_query: Optional[query.AxisQuery] = None,
```
atolopko-czi, line 51: docstring for this method indicates `See :class:`query.ExperimentAxisQuery` for details on usage`.  But in fact I think the problem will still persist in `somacore` code.  If caller of this methods specifies `None`, somacore will not convert that to AxisQuery().  Does the fix need to be pushed down into somacore perhaps? Either way, suggest adding a test in this repo to ensure correct behavior.

atolopko-czi (changes requested):
A concern noted in comments.

bkmartinjr (approved):
this looks great - ty

bkmartinjr:
> LGTM. Should we also add a corresponding unit test in TileDB-SOMA repo?

_Please_ add a test - the lack of them was why we missed this regression previously.

PR #100, opened 2023-01-25
Add `close` method and context-manager operations

thetorpedodog:
This adds a `close` method to SOMAObject, which allows implementations to specify cleanup operations, and hooks it up to the context manager protocol by default.

Also cleans up other delete callsites to do our own delete work *before* calling the superclass delete.


bkmartinjr (approved):
LGTM

johnkerl (approved):

PR #101, opened 2023-01-25
Empty query results typing

bkmartinjr:
Several small changes to ensure that empty query results have types that are compatible with AnnData, avoiding warnings/errors from AnnData.

1. The empty matrix result in the fast CSR builder was creating an empty float64 matrix.  AnnData has a strong preference for float32, and will issue a warning on anything else.  As the actual type is relatively immaterial (it is an empty matrix), this PR changes to float32 to minimize complaints from AnnData.
2. The axis dataframes need to have an index type of str.  Previous code did not work correctly when the dataframe was empty.
3. When the resulting X is empty (nnz==0), need to ensure it has the same shape as the axis.

Fixes: single-cell-data/TileDB-SOMA#805

### DIFF
--- a/python-spec/src/somacore/query/_fast_csr.py
+++ b/python-spec/src/somacore/query/_fast_csr.py
@@ -117,8 +117,10 @@ class _CSRAccumulator:
         nnz = sum(len(chunk[2]) for chunk in self.coo_chunks)
         index_dtype = _select_dtype(nnz)
         if nnz == 0:
-            # no way to infer matrix dtype, so use default and return empty matrix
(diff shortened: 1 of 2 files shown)


thetorpedodog (approved):

bkmartinjr:
@thetorpedodog - I forgot one small change which I pushed after your review.  Mentioning in case you want to take another look.

thetorpedodog:
> @thetorpedodog - I forgot one small change which I pushed after your review. Mentioning in case you want to take another look.

I saw that change roll in as I was looking at it; it’s good.

Issue #45, opened 2022-10-19
X arrays can only be sparse


johnkerl:
From recent conversation with @bkmartinjr I  believe `X` arrays will always be sparse.

If I am correct, this should be updated at https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#somaexperiment--somameasurement.

cc @thetorpedodog and @gspowley.

bkmartinjr:
Hi @johnkerl - I apologize for adding noise to the system, but after reviewing my notes, I concluded that we decided to allow both dense & sparse.  Sparse is far more common for transcriptomic data, but there are use cases for both.

The prose spec is correct (checked it). The only bug I saw is image/diagram, which I will fix. I will leave this bug assigned to me until that is done.  Thanks for following up on this!

johnkerl:
X arrays can now be sparse or dense. See also #106 which did related work.

PR #106, opened 2023-01-27
Remove obs_ms and var_ms from images/SOMAExperiment.png

### LINKED ISSUE #63: Spec should not contain obs_ms/var_ms
These were deleted, but not removed from the diagram/image.  Needs to be consistent, by cleaning up the image/diagram.

johnkerl:
Fixes #63

See attached image -- `obs_ms` and `var_ms` are no longer in the verbal spec.


bkmartinjr (changes requested):
the types documented in the X layer are wrong - per the spec, X layers can be Dense or Sparse.

See #45.

johnkerl:
@bkmartinjr please share the original document you used to create this image.

I was photo-editing, which is non-optimal.

bkmartinjr:
> @bkmartinjr please share the original document you used to create this image.

will do offline as it is a gdoc thingy.  It was previously shared with @aaronwolen

we may want to consider some other master - but I don't have a good suggestion for an alternative

johnkerl:
@maniarathi ping

atolopko-czi (approved):
The desired sections have been removed. But honestly, the larger resolution image is being downscaled in the README.md and is no longer displayed at its native resolution. While a minor concern, this is an important image in the spec, and should look better, IMO. Can we show at 1:1 resolution? Too big?

I'll approve as is, but suggest we create a new issue to recreate the image from its original image editing app  at the desired resolution (understanding we need to figure out where that is).

Github visual diff didn't work, so I diffed between commits https://github.com/single-cell-data/SOMA/blob/2f090afd1b788f115407b95fed54a3de756e07fb/images/SOMAExperiment.png and https://github.com/single-cell-data/SOMA/blob/8a2e014beecb648de00227974c0f06e20d5da417/images/SOMAExperiment.png

johnkerl:
@atolopko-czi I'll downsize as a follow-on PR -- thanks, good catch!! :)

PR #108, opened 2023-02-01
Explicitly use object-identity–based equality for SOMA objects.

thetorpedodog:
When inheriting from a type that defines an `__eq__` operation, that `__eq__` will override the built-in `__eq__`. In the case of the `MutableMapping` type, that means that it will try to use element-wise comparison against any other mapping type. This is effectively an attractive nuisance. Instead, we explicitly use the default comparison, which is based on object identity rather than contents or anything.


atolopko-czi (approved):
LGTM

PR #109, opened 2023-02-01
Pull "ephemeral" collections into their own module.

thetorpedodog:
The meaning and usage of the various `SimpleWhatever` collections was unclear with them colocated with the abstract implementation. Moving them to their own `ephemeral` module and not having them in the main `somacore` namespace will make them less attractive. They still have value in being the most basic possible implementation of a collection, and are useful in tests for this package.

In the process of removing the `SimpleCollection` export from the package root, also changes the `__init__.py` to just import module members rather than the module itself. Since this is at the top level and is exclusively for re-export, this does not have the same potential circular dependency problems as doing so within internal modules, while reducing repetition and the possibility of errors.

### DIFF
--- a/python-spec/src/somacore/ephemeral/collections.py
+++ b/python-spec/src/somacore/ephemeral/collections.py
@@ -31,7 +31,7 @@ class Collection(collection.Collection[_ST]):
 
     @property
     def uri(self) -> str:
-        return f"somacore:simple-collection:{id(self):x}"
+        return f"somacore:ephemeral-collection:{id(self):x}"
 
     @property
     def metadata(self) -> Dict[str, Any]:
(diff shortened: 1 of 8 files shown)


python-spec/src/somacore/ephemeral/collections.py
```diff
@@ -0,0 +1,88 @@
...
+        """Creates a new Collection.
+
+        Arguments and kwargs are provided as in the ``dict`` constructor.
+        """
+        self._entries: Dict[str, _ST] = dict(*args, **kwargs)
+        self._metadata: Dict[str, Any] = {}
+
+    @property
+    def uri(self) -> str:
+        return f"somacore:simple-collection:{id(self):x}"
```
atolopko-czi, line 34: `ephemeral-collection` for consistency?

python-spec/testing/test_collection.py
```diff
@@ -1,18 +1,16 @@
...
-from somacore import collection
-from somacore import experiment
-from somacore import measurement
+from somacore import ephemeral
 
 
-class SimpleCollectionTest(unittest.TestCase):
+class EphemeralCollectionTest(unittest.TestCase):
     def test_basic(self):
         # Since the SimpleCollection implementation is straightforward this is
```
atolopko-czi, line 9: might want to `s/\bSimple(.+)\b/Ephemeral$1/`

atolopko-czi (approved):
LGTM. The separation helps!

Just a couple spots where "simple" still exists; suggest a full search & replace. Approving optimistically.

python-spec/testing/test_collection.py
thetorpedodog: Whoops! Thank you.

python-spec/src/somacore/ephemeral/collections.py
thetorpedodog: Also fixed.

PR #114, opened 2023-02-03
Add support for str and bytes in AxisQuery

thetorpedodog:
Because our backend now supports string and bytes queries [1], we no longer need to reject them at the input stage. Also adds more tests.

[1]: https://github.com/single-cell-data/TileDB-SOMA/pull/861

---

Related to: https://github.com/single-cell-data/TileDB-SOMA/issues/419

### DIFF
--- a/python-spec/src/somacore/query/axis.py
+++ b/python-spec/src/somacore/query/axis.py
@@ -18,6 +18,10 @@ def _canonicalize_coords(
     """
     if in_coords is None:
         return (slice(None),)
+    if not isinstance(in_coords, Sequence):
+        raise TypeError(
+            f"query coordinates must be a sequence, not a single {type(in_coords)}"
+        )
     if not _is_normal_sequence(in_coords):
         raise TypeError(
             "query coordinates must be a normal sequence, not `str` or `bytes`."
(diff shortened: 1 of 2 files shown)


python-spec/src/somacore/query/axis.py
```diff
@@ -28,10 +32,10 @@ def _canonicalize_coords(
 def _canonicalize_coord(coord: options.SparseDFCoord) -> options.SparseDFCoord:
     """Validates a single coordinate, freezing mutable sequences."""
     if coord is None or isinstance(
-        coord, (int, slice, pa.Array, pa.ChunkedArray, np.ndarray)
+        coord, (bytes, int, slice, str, pa.Array, pa.ChunkedArray, np.ndarray)
```
johnkerl, line 35: nit: put `bytes` next to `str`

johnkerl (approved):
✨

python-spec/src/somacore/query/axis.py
thetorpedodog: I considered that, but decided on this since it puts all the built-in types in alphabetical order.

Issue #107, opened 2023-01-31
Document stateful API design


Shelnutt2:
Now that #44 has been accepted and implemented for a stateful API, we need to add documentation about the design and APIs.

PR #118, opened 2023-02-09
[spec] Specify object lifecycle and related operations

thetorpedodog:
This formalizes the object lifecycle and the various creation, opening, and closing methods we are already using in the tiledbsoma Python implementation.

See #107.

### DIFF
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -415,7 +415,7 @@ In addition, SOMACollection supports operations to manage the contents of the co
 
 A SOMA collection also manages the lifecycle of objects directly instantiated by it.
 Objects accessed via getting a Collection element, or objects created with one of the <code>add*new*<var>object_type</var></code> methods are considered "owned" by the collection.
-This allows all the objects to be automatically closed together by [the collection's close operation](#operation-close-collection-types).
+All such objects will be automatically closed together by [the collection's close operation](#operation-close-collection-types).
 
 ### Operation: create() (Collection types)
 
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -124,13 +124,16 @@ SOMACollection is an unordered, `string`-keyed map of values. Values may be any 
 
 #### Collection entry URIs
 
-Collections refer to their components by URI.
+Collections refer to their elements by URI.
 A collection may store its reference to an element by **absolute** URI or **relative** URI.
 
 A reference to an **absolute** URI is, as the name implies, absolute.
+This means that the collection stores the full URI of the element.
 When the backing storage of the collection itself is moved, absolute URI references do not change.
 
 A collection can also refer to sub-paths of itself by **relative** URI.
+This means that the collection's elements are stored as suffixes of the collection's path; for instance for a collection at `backend://host/dataset`, the name `child` would refer to `backend://host/dataset/child`.
+This applies both to truly hierarchical systems (like POSIX filesystems) and systems that are technically non-hierarchical but can emulate a hierarchical namespace, like Amazon S3 or Google Cloud Storage.
 If the entire directory storing a collection and a child referred to by relative URI is moved, the relative URI references now refer to the changed base URI.
 While the term "relative" is used, only **child** URIs are supported as relative URIs; relative paths like `../sibling-path/sub-entry` are not supported.
 
@@ -176,11 +179,11 @@ When `file:///soma-dataset/old-data/the-collection` is opened, the URIs will be 
 - `absolute`: `file:///soma-dataset/collection/abspath`
   (This resolved URI is the same as before. However, the data is no longer at this location; attempting to access this element will fail.)
 - `external-absolute`: `file:///soma-dataset/more-data/other-data`
-  (This URI still points to the same data.)
+  (This URI is identical and still points to the same data.)
 - `relative`: `file:///soma-dataset/old-data/the-collection/relpath`
   (This URI resolves differently and still points to the same data even after it has been moved.)
 
-In general, absolute and relative URIs can both be used interchangeably.
+In general, absolute and relative URIs can both be used interchangeably within the same collection (as in the example above).
 However, in certain situations, an implementation may support only absolute or relative URIs.
 For instance, a purely filesystem-based implementation may support only relative URIs, or a non-hierarchichal storage backend may support only absolute URIs.
 
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -122,6 +122,68 @@ SOMACollection is an unordered, `string`-keyed map of values. Values may be any 
 - as a container of independent objects (e.g., a collection of single-cell datasets, each manifest as a [SOMAExperiment](#soma-experiment) object)
 - as the basis for building other composed types (e.g., using SOMACollection to organize pre-defined fields in [SOMAExperiment](#soma-experiment) such as multiple layers of `X`).
 
+#### Collection entry URIs
+
+Collections refer to their components by URI.
+A collection may store its reference to an element by **absolute** URI or **relative** URI.
+
+A reference to an **absolute** URI is, as the name implies, absolute.
+When the backing storage of the collection itself is moved, absolute URI references do not change.
+
+A collection can also refer to sub-paths of itself by **relative** URI.
+If the entire directory storing a collection and a child referred to by relative URI is moved, the relative URI references now refer to the changed base URI.
+While the term "relative" is used, only **child** URIs are supported as relative URIs; relative paths like `../sibling-path/sub-entry` are not supported.
+
+Consider a directory tree of a SOMA implementation that uses a file named `.collection-info` to store the URIs of the collection's contents:
+
+- `file:///soma-dataset`
+  - `collection`
+    - `.collection-info`
+      - `absolute`: `file:///soma-dataset/collection/abspath`
+      - `external-absolute`: `file:///soma-dataset/more-data/other-data`
+      - `relative`: `relpath`
+    - `abspath`
+      - [contents]
+    - `relpath`
+      - [contents]
+  - `more-data`
+    - [contents]
+
+When `file:///soma-dataset/collection` is opened, the URIs will be resolved as follows:
+
+- `absolute`: `file:///soma-dataset/collection/abspath`
+- `external-absolute`: `file:///soma-dataset/more-data/other-data`
+- `relative`: `file:///soma-dataset/collection/relpath`
+
+If the entire `collection` directory is moved to a new path:
+
+- `file:///soma-dataset`
+  - `old-data`
+    - `the-collection`
+      - `.collection-info`
+        - `absolute`: `file:///soma-dataset/collection/abspath`
+        - `external-absolute`: `file:///soma-dataset/more-data/other-data`
+        - `relative`: `relpath`
+      - `abspath`
+        - [contents]
+      - `relpath`
+        - [contents]
+  - `more-data`
+    - [contents]
+
+When `file:///soma-dataset/old-data/the-collection` is opened, the URIs will be resolved as follows:
+
+- `absolute`: `file:///soma-dataset/collection/abspath`
+  (This resolved URI is the same as before. However, the data is no longer at this location; attempting to access this element will fail.)
+- `external-absolute`: `file:///soma-dataset/more-data/other-data`
+  (This URI still points to the same data.)
+- `relative`: `file:///soma-dataset/old-data/the-collection/relpath`
+  (This URI resolves differently and still points to the same data even after it has been moved.)
+
+In general, absolute and relative URIs can both be used interchangeably.
+However, in certain situations, an implementation may support only absolute or relative URIs.
+For instance, a purely filesystem-based implementation may support only relative URIs, or a non-hierarchichal storage backend may support only absolute URIs.
+
 ### SOMADataFrame
 
 `SOMADataFrame` is a multi-column table with a user-defined schema, defining the number of columns and their respective column name and value type. The schema is expressed as an Arrow `Schema`.
@@ -236,7 +298,7 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
 ## Object lifecycle
 
 SOMA object instances have a lifecycle analogous to the lifecycle of OS-level file objects.
-A SOMA object is instantiated by **opening** a provided URI, in one of _read_ (`r`) or _write_ (`w`) mode.
+A SOMA object is instantiated by **opening** a provided URI, in one of [_read_ or _write_ mode](#openmode).
 The opened object can be manipulated (pursuant to its open state) until it is **closed**, at which point any further calls which attempt to access or modifying backing data will fail (again, akin to a file object).
 
 SOMA objects are open for **exclusively** reading or writing.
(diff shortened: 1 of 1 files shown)


bkmartinjr (dismissed):
Caveat: I did a very fast review.  But based upon that, it looks good to me.

abstract_specification.md
```diff
@@ -233,68 +233,290 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+| del(string key)                         | Remove the key/value from the collection. Does not delete the underlying object (value). |
+| add_new_collection(string key, ...)     | Creates a new sub-Collection and adds it to this Collection.                             |
+| add_new_dataframe(string key, ...)      | Creates a new DataFrame and adds it to this Collection.                                  |
+| add_new_dense_ndarray(string key, ...)  | Creates a new DenseNDArray and adds it to this Collection.                               |
+| add_new_sparse_ndarray(string key, ...) | Creates a new SparseNDArray and adds it to this Collection.                              |
+
+A SOMA collection also manages the lifecycle of objects directly instantiated by it.
+Objects accessed via getting a Collection element, or objects created with one of the <code>add*new*<var>object_type</var></code> methods are considered "owned" by the collection.
+This allows all the objects to be automatically closed together by [the collection's close operation](#operation-close-collection-types).
```
mlin, line 418:
```suggestion
All such objects will be automatically closed together by [the collection's close operation](#operation-close-collection-types).
```

mlin (approved):
LGTM, one minor language tightening suggested inline

abstract_specification.md
thetorpedodog: This is good.

abstract_specification.md
```diff
@@ -233,68 +233,290 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
-| exists(uri) -> bool                     | Return true if object exists and is a SOMADataFrame.                           |
-| get soma_type                           | Returns the constant "SOMADataFrame"                                           |
-| get schema -> Arrow.Schema              | Return data schema, in the form of an Arrow Schema                             |
-| get index_column_names -> [string, ...] | Return index (dimension) column names.                                         |
-| get count -> int                        | Return the number of rows in the SOMADataFrame.                                |
-| read                                    | Read a subset of data from the SOMADataFrame                                   |
-| write                                   | Write a subset of data to the SOMADataFrame                                    |
+| Operation                                | Description                                        |
+| ---------------------------------------- | -------------------------------------------------- |
+| static create(uri, ...) -> SOMADataFrame | Create a SOMADataFrame.                            |
```
johnkerl, line 513: ditto re end w/ period or not

johnkerl (changes requested):

abstract_specification.md
```diff
@@ -233,68 +233,290 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+SOMA objects are open for **exclusively** reading or writing.
+When a SOMA object is open for writing, only metadata, the schema, and information directly derived from the schema may be read.
+Additionally, for Collection objects, the members of the collection are also accessible when the collection is open in write mode.
+For example:
+
+```
+writable_collection = soma_impl.open('file:///some/collection', 'w')
+dataframe = writable_collection['dataframe']  # OK
+dataframe.schema  # OK
+dataframe.read(...)  # This is NOT allowed and will cause an error.
```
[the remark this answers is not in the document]
bkmartinjr: For clarity, our internal specs tend to always use the [RFC2119](https://www.ietf.org/rfc/rfc2119.txt) standard (i.e., consistently use `NOT`), but given that this doc has not already adopted it, that is probably an aspirational change.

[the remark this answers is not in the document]
thetorpedodog: an author [OUGHT TO](https://www.rfc-editor.org/rfc/rfc6919#section-4) use the caps consistently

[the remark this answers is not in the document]
johnkerl: @bkmartinjr TIL about RFC2119! :) 

If we want to go that route, sure.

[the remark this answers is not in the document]
bkmartinjr: there are very good reasons for it - which is why the IETF uses it (as does many other orgs like W3C).  It is a future decision which should not hold up this PR -- but coincidentally the topic was it was just raised inside CZI (regarding this very doc being unclear).  

[Here is](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md) an example of one of our specs that strongly benefits from it.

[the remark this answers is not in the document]
johnkerl: I filed #122 -- this is awesome info -- thanks @bkmartinjr !! :)

thetorpedodog (commented):
lots of changes coming in hot

abstract_specification.md
```diff
@@ -233,68 +233,290 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+class Collection:
+    @classmethod
+    def open(cls, uri, ...): ...
+```
+
+```
+soma_collection_t soma_collection_create(char *uri, ...) { ... }
+```
+
+Where possible, the lifecycle should be integrated into language-specific resource management frameworks (e.g. context management with `with` in Python, try-with-resources in Java).
```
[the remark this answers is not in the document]
johnkerl: I continue to disagree but won't block the PR.

johnkerl (approved):

thetorpedodog:
@bkmartinjr @mlin actually I would appreciate if you all took another look particularly for the description of URI semantics that I added.

@johnkerl I clicked you by accident but if you want to give it another ✅ be my guest

abstract_specification.md
```diff
@@ -122,6 +122,68 @@ SOMACollection is an unordered, `string`-keyed map of values. Values may be any
...
+#### Collection entry URIs
+
+Collections refer to their components by URI.
+A collection may store its reference to an element by **absolute** URI or **relative** URI.
+
+A reference to an **absolute** URI is, as the name implies, absolute.
+When the backing storage of the collection itself is moved, absolute URI references do not change.
+
+A collection can also refer to sub-paths of itself by **relative** URI.
+If the entire directory storing a collection and a child referred to by relative URI is moved, the relative URI references now refer to the changed base URI.
```
bkmartinjr, line 134: you might add some wording here that points out that "relative" is an attribute of the particular storage system, eg. S3 really doesn't have directories, but has path prefixes that serve the same purpose.  TileDB-Cloud shows another bunch of special cases if I understand it right.

So I would point out that relative is optional.

And I would reword this to something like:    Where relative URIs are supported by the storage system, collections referencing children by a relative URI will always resolve the child with respect to the parent (base) URI.  For example, on a Posix file system, if the entire directory storing a collection .... <same as you have>

abstract_specification.md
```diff
@@ -233,68 +233,290 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+### Operation: close (Collection types)
+
+In addition to the details of the [common close operation](#operation-close), there are some semantics specific to collection types.
+
+When a SOMA collection is directly used to instantiate child SOMA objects, those SOMA objects are considered "owned" by the collection, and the collection manages their lifecycle.
+When the collection is closed, it must close any of these child objects before closing itself.
+For example:
+
+```
+root = soma_impl.open('path://netloc/root/collection', 'w')
```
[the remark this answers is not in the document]
bkmartinjr: we should eventually move to an IDL-like definition if we want the spec to be durable.  But as John says, that is future work, not immediately needed (IMHO)

bkmartinjr (approved):
One suggestion for a clarification around relative URI resolution in collections.  Otherwise LGTM

abstract_specification.md
thetorpedodog: Added some description of this, along with language indicating that we treat “fake” hierarchies just like normal hierarchies.

johnkerl (approved):

thetorpedodog:
@atolopko-czi @aaronwolen since this is a spec change we do want unanimous approvals on this so please take a look. Thanks!

johnkerl:
@thetorpedodog ✅  ✅

aaronwolen (approved):
This is really great

abstract_specification.md
```diff
@@ -233,68 +298,298 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+| del(string key)                                     | Remove the key/value from the collection. Does not delete the underlying object (value). |
+| add_new_collection(string key, ...)                 | Creates a new sub-Collection and adds it to this Collection.                             |
+| add_new_dataframe(string key, ...)                  | Creates a new DataFrame and adds it to this Collection.                                  |
+| add_new_dense_ndarray(string key, ...)              | Creates a new DenseNDArray and adds it to this Collection.                               |
+| add_new_sparse_ndarray(string key, ...)             | Creates a new SparseNDArray and adds it to this Collection.                              |
+
+A SOMA collection also manages the lifecycle of objects directly instantiated by it.
+Objects accessed via getting a Collection element, or objects created with one of the <code>add_new\_<var>object_type</var></code> methods are considered "owned" by the collection.
+All such objects will be automatically closed together by [the collection's close operation](#operation-close-collection-types).
```
atolopko-czi, line 483: Also worth mentioning they inherit the parent's context? Or do we consider that implementation-specific?

abstract_specification.md
```diff
@@ -233,68 +298,298 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+# external_object remains open because it is not owned by root.
+```
+
+A user may close a child object before closing the parent object.
+Because `close` is idempotent, this is explicitly allowed.
+Closing the child will _not_ otherwise affect the parent object; it will remain open as usual.
+
+### Operation: set
+
+Adds an entry to the collection, overwriting an existing key if one exists already.
```
atolopko-czi, line 542: Does the spec have an opinion on what happens to data originally stored by the object that is replaced at a specified key? Is it orphaned or removed?

abstract_specification.md
```diff
@@ -233,68 +298,298 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+- key - The key to set.
+- value - The value to set the key to. If a user sets a collection entry to a type inconsistent with the type of that key (e.g. a defined key in an Experiment) or with the general type of the collection (e.g. a collection of Measurements), behavior is unspecified. Implementations are encouraged to raise an error if possible, to prevent latent errors when later attempting to use the data.
+- uri_type - How the collection should refer to the URI of the newly-added element, whether by absolute or relative URI. The default is `auto`, which will use a relative URI if possible but otherwise use an absolute URI. If `absolute`, the entry will always use an absolute URI. If `relative`, the entry will always use a relative URI, and an error will be raised if a relative URI cannot be used.
+
+### Operation: add_new\_<var>object_type</var>
+
+Each <code>add_new\_<var>object_type</var></code> method creates a new SOMA dataset in storage, adds it to the collection, and returns it to the user. The newly-created entry has the same `context` value as the existing collection and is [owned by the current collection](#operation-close-collection-types).
```
atolopko-czi, line 556:
```suggestion
Each <code>add_new\_<var>object_type</var></code> method creates a new SOMA dataset in storage, adds it to the collection (with the same semantics as the `create` operation), and returns it to the user. The newly-created entry has the same `context` value as the existing collection and is [owned by the current collection](#operation-close-collection-types).
```

abstract_specification.md
```diff
@@ -233,68 +298,298 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+```
+add_new_dataframe(string key, string uri = "", ...) -> SOMADataFrame
+add_new_dense_ndarray(string key, string uri = "", ...) -> SOMADenseNDArray
+add_new_sparse_ndarray(string key, string uri = "", ...) -> SOMASparseNDArray
+```
+
+Parameters:
+
+- key: The key to add the new element at. This cannot already be a key of the collection.
+- uri: An optional parameter to specify a URI to create the new collection, which may be [relative or absolute](#collection-entry-uris). If the URI is relative, the new entry will be added with that relative URI. If the URI is absolute, the new entry will be added with that absolute URI. If a collection already exists at the user-provided URI, the operation should fail. If the user does not specify a URI, the collection will generate a new URI for the entry. When possible, this should be a relative URI based on a sanitized version of the key.
```
atolopko-czi, line 567: Do this need to mention that relative URIs cannot be "multi-level"? I think that's what the spec mentions above.

abstract_specification.md
```diff
@@ -233,68 +298,298 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
+- The remaining parameters are passed directly to the respective type's `create` static method, except for `context`, which is always set to the current collection's context.
+
+`add_new_collection` has an extra parameter allowing control over the type of collection added:
+
+```
+add_new_collection(string key, CollectionType type, string uri = "", PlatformConfig platform_config) -> SOMACollection
```
atolopko-czi, line 573: Completely out-of-scope for this PR, but I'm now just wondering why we didn't consider having explicit methods for the various types of collections? `add_new_experiment`, `add_new_measurement`, `add_new_collection`.

abstract_specification.md
```diff
@@ -233,68 +298,298 @@ Any given storage "engine" upon which SOMA is implemented may have additional fe
...
-| read                                    | Read a subset of data from the SOMADataFrame                                   |
-| write                                   | Write a subset of data to the SOMADataFrame                                    |
+| Operation                                | Description                                         |
+| ---------------------------------------- | --------------------------------------------------- |
+| static create(uri, ...) -> SOMADataFrame | Create a SOMADataFrame.                             |
+| get soma_type                            | Returns the constant "SOMADataFrame".               |
+| get schema -> Arrow.Schema               | Return data schema, in the form of an Arrow Schema. |
+| get index_column_names -> [string, ...]  | Return index (dimension) column names.              |
+| get count -> int                         | Return the number of rows in the SOMADataFrame.     |
+| read                                     | Read a subset of data from the SOMADataFrame.       |
```
atolopko-czi, line 591: `Read a subset of data iteratively from the SOMADataFrame`?  Whether or not one is reading a subset, the iterative nature of reads seems like a core part of the design, even at the spec level, given the returns type for DataFrame.read (and SparseNDArray.read())

While most people bothering to read a spec like this will understand that a subset includes the whole set, do we want to rely upon reader knowing that? "Read all or a subset..."

abstract_specification.md
```diff
@@ -465,26 +763,24 @@ Parameters:
...
+| static create(uri, ...) -> SOMASparseNDArray | Create a SOMASparseNDArray named with the URI.                          |
+| get soma_type                                | Returns the constant "SOMASparseNDArray".                               |
+| get shape -> (int, ...)                      | Return length of each dimension, always a list of length `ndim`.        |
+| get ndim -> int                              | Return number of dimensions.                                            |
+| get schema -> Arrow.Schema                   | Return data schema, in the form of an Arrow Schema.                     |
+| get is_sparse -> True                        | Return the constant True.                                               |
+| get nnz -> uint                              | Return the number stored values in the array, including explicit zeros. |
+| read                                         | Read a slice of data from the SOMASparseNDArray.                        |
+| write                                        | Write a slice of data to the SOMASparseNDArray.                         |
```
atolopko-czi, line 776: important to use `slice` here instead of `subset` as we do above? If it's a technical term we explicitly want, ok.

abstract_specification.md
```diff
@@ -608,6 +948,7 @@ Array read operations can return results in a variety of formats. The `SOMABatch
 Summary:
 
 ```
+open(string uri, ...) -> SOMA object      # identify a SOMA object and open it
```
atolopko-czi, line 951:
```suggestion
open(string uri, ...) -> SOMAObject      # identify a SOMA object and open it
```
?

atolopko-czi (approved):
Really great to read the official details on open/close, read/write, etc.! A number of very minor questions and suggestions. Approving as is, since I found no blockers, but have a look at my comments before merging.

mlin (approved):

abstract_specification.md
```diff
@@ -122,6 +122,71 @@ SOMACollection is an unordered, `string`-keyed map of values. Values may be any
...
+While the term "relative" is used, only **child** URIs are supported as relative URIs; relative paths like `../sibling-path/sub-entry` are not supported.
+
+Consider a directory tree of a SOMA implementation that uses a file named `.collection-info` to store the URIs of the collection's contents:
+
+- `file:///soma-dataset`
+  - `collection`
+    - `.collection-info`
+      - `absolute`: `file:///soma-dataset/collection/abspath`
+      - `external-absolute`: `file:///soma-dataset/more-data/other-data`
+      - `relative`: `relpath`
```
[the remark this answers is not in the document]
thetorpedodog: Updated to represent `.collection-info` as a real file and present its contents differently from other paths.

PR #119, opened 2023-02-09
[python] Add missing lifecycle methods and expand docstrings

thetorpedodog:
- Adds `mode`, `close`, and `closed` methods/properties.
- Fills in most of the docstrings with details about parameters and expected behavior.

---

This change depends upon the result of https://github.com/single-cell-data/TileDB-SOMA/pull/910.

Part of #107 and https://github.com/single-cell-data/TileDB-SOMA/issues/638.

### DIFF
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -27,7 +27,14 @@ class SOMAObject(metaclass=abc.ABCMeta):
         context: Optional[Any] = None,
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> Self:
-        """Opens the SOMA object at the given URL."""
+        """Opens the SOMA object of this type at the given URI.
+
+        :param uri: The URI of the object to open.
+        :param mode: The mode to open this in, either `r` or `w`.
+        :param context: The Context value to use when opening the object.
+        :param platform_config: Platform configuration options specific to
+            this open operation.
+        """
         raise NotImplementedError()
 
     @property
@@ -59,6 +66,18 @@ class SOMAObject(metaclass=abc.ABCMeta):
         and writes to it (provided the object is opened) are reflected in
         storage.
         """
+        raise NotImplementedError()
+
+    @property
+    @abc.abstractmethod
+    def mode(self) -> options.OpenMode:
+        """Returns the mode this object was opened in."""
+        raise NotImplementedError()
+
+    @property
+    @abc.abstractmethod
+    def closed(self) -> bool:
+        """Returns True if this object has been closed; False if still open."""
         raise NotImplementedError()
 
     soma_type: ClassVar[LiteralString]
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -13,7 +13,7 @@ from . import options
 
 
 class SOMAObject(metaclass=abc.ABCMeta):
-    """A sentinel interface indicating that this type is a SOMA object."""
+    """The base type for all SOMA objects, containing common behaviors."""
 
     __slots__ = ("__weakref__",)
 
@@ -28,6 +28,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> Self:
         """Opens the SOMA object of this type at the given URI.
+        [lifecycle: experimental]
 
         :param uri: The URI of the object to open.
         :param mode: The mode to open this in, either `r` or `w`.
@@ -40,12 +41,13 @@ class SOMAObject(metaclass=abc.ABCMeta):
     @property
     @abc.abstractmethod
     def uri(self) -> str:
-        """Returns the URI of this SOMA object."""
+        """Returns the URI of this SOMA object. [lifecycle: experimental]"""
         raise NotImplementedError()
 
     @property
     def context(self) -> Any:
         """A value storing implementation-specific configuration information.
+        [lifecycle: experimental]
 
         This contains long-lived (i.e., not call-specific) information that is
         used by the SOMA implementation to access storage. This may include
@@ -60,7 +62,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
     @property
     @abc.abstractmethod
     def metadata(self) -> MutableMapping[str, Any]:
-        """The metadata of this SOMA object.
+        """The metadata of this SOMA object. [lifecycle: experimental]
 
         The returned value directly references the stored metadata; reads from
         and writes to it (provided the object is opened) are reflected in
@@ -71,13 +73,17 @@ class SOMAObject(metaclass=abc.ABCMeta):
     @property
     @abc.abstractmethod
     def mode(self) -> options.OpenMode:
-        """Returns the mode this object was opened in."""
+        """Returns the mode this object was opened in, either ``r`` or ``w``.
+        [lifecycle: experimental]
+        """
         raise NotImplementedError()
 
     @property
     @abc.abstractmethod
     def closed(self) -> bool:
-        """Returns True if this object has been closed; False if still open."""
+        """Returns True if this object has been closed; False if still open.
+        [lifecycle: experimental]
+        """
         raise NotImplementedError()
 
     soma_type: ClassVar[LiteralString]
@@ -93,6 +99,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
 
     def close(self) -> None:
         """Releases any external resources held by this object.
+        [lifecycle: experimental]
 
         An implementation of close must be idempotent.
         """
--- a/python-spec/src/somacore/data.py
+++ b/python-spec/src/somacore/data.py
@@ -28,7 +28,9 @@ _RO_AUTO = options.ResultOrder.AUTO
 
 
 class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
-    """A multi-column table with a user-defined schema."""
+    """A multi-column table with a user-defined schema.
+    [lifecycle: experimental]
+    """
 
     __slots__ = ()
     soma_type: Final = "SOMADataFrame"  # type: ignore[misc]
@@ -47,6 +49,13 @@ class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
         context: Optional[Any] = None,
     ) -> Self:
         """Creates a new ``DataFrame`` at the given URI.
+        [lifecycle: experimental]
+
+        The schema of the created dataframe will include a column named
+        ``soma_joinid`` of type ``pyarrow.int64``. If a ``soma_joinid`` column
+        is present in the provided schema, it must be of the correct type. If no
+        ``soma_joinid`` column is provided, one will be added. This column may
+        be used as an indexed column.
 
         :param uri: The URI where the ``DataFrame`` will be created.
         :param schema: Arrow schema defining the per-column schema. This schema
@@ -75,6 +84,7 @@ class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> "ReadIter[pa.Table]":
         """Reads a user-defined slice of data into Arrow tables.
+        [lifecycle: experimental]
 
         :param coords: for each index dimension, which rows to read.
             Defaults to ``()``, meaning no constraint -- all IDs.
@@ -109,13 +119,16 @@ class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
         - Coordinates can be specified as a scalar value, a Python sequence
           (``list``, ``tuple``, etc.), a NumPy ndarray, an Arrow array, or
           similar objects (as defined by ``SparseDFCoords``).
-        - Slices are doubly inclusive: ``slice(2, 4)`` means ``[2, 3, 4]``,
-          not ``[2, 3]``.  Slice *steps* may not be used: ``slice(10, 20, 2)``
-          is invalid.  ``slice(None)`` places no constraint on the dimension.
-          Half-specified slices like ``slice(None, 99)`` and ``slice(5, None)``
-          specify all indices up to and including that value, and all indices
+        - Slices specify a closed range: ``slice(2, 4)`` includes both 2 and 4.
+          Slice *steps* may not be used: ``slice(10, 20, 2)`` is invalid.
+          ``slice(None)`` places no constraint on the dimension. Half-specified
+          slices like ``slice(None, 99)`` and ``slice(5, None)`` specify
+          all indices up to and including the value, and all indices
           starting from and including the value.
-        - Negative indexing is not supported.
+        - Negative values in indices and slices are treated as raw domain values
+          and not as indices relative to the end, unlike traditional Python
+          sequence indexing. For instance, ``slice(-10, 3)`` indicates the range
+          from −10 to 3 on the given dimension.
         """
         raise NotImplementedError()
 
@@ -127,11 +140,10 @@ class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> None:
         """Writes the data from an Arrow table to the persistent object.
+        [lifecycle: experimental]
 
         As duplicate index values are not allowed, index values already present
         in the object are overwritten and new index values are added.
-
-        [lifecycle: experimental]
 
         :param values: An Arrow table containing all columns, including
             the index columns. The schema for the values must match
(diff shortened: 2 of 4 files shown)


python-spec/src/somacore/data.py
```diff
@@ -65,7 +76,48 @@ def read(
     ) -> "ReadIter[pa.Table]":
         """Reads a user-defined slice of data into Arrow tables.
 
-        TODO: Further per-param documentation.
+        :param coords: for each index dimension, which rows to read.
+            Defaults to ``None``, meaning no constraint -- all IDs.
+        :param column_names: the named columns to read and return.
+            Defaults to ``None``, meaning no constraint -- all column names.
+        :param partitions: an optional ``ReadPartitions`` hint to indicate
```
johnkerl, line 83: here you have backticked

python-spec/src/somacore/data.py
```diff
@@ -46,7 +46,18 @@ def create(
         platform_config: Optional[options.PlatformConfig] = None,
         context: Optional[Any] = None,
     ) -> Self:
-        """Creates a new DataFrame."""
+        """Creates a new DataFrame at the given URI.
+
+        :param uri: The URI where the DataFrame will be created.
```
johnkerl, line 51: here you have non-backticked -- let's be consistent and backtick code names

python-spec/src/somacore/data.py
```diff
@@ -65,7 +76,48 @@ def read(
...
-        TODO: Further per-param documentation.
+        :param coords: for each index dimension, which rows to read.
+            Defaults to ``None``, meaning no constraint -- all IDs.
+        :param column_names: the named columns to read and return.
+            Defaults to ``None``, meaning no constraint -- all column names.
+        :param partitions: an optional ``ReadPartitions`` hint to indicate
+            how results should be organized.
+        :param result_order: order of read results.
+            This can be one of 'row-major', 'col-major', or 'auto'.
+        :param value_filter: an optional [value filter] to apply to the results.
```
johnkerl, line 87: needs an example, no matter how trivial -- or, a statement that the query language is implementation-dependent

this is super-awkward now that this _code_ in the _spec_ repo now appears as on-line help for the _impl_ repo ......

we've had consistent feedback over the last few months on this project "what's an example of a value-filter query?" -- and now we're in a spot where we've made it difficult to satisfy that legitimate user ask ...

python-spec/src/somacore/data.py
```diff
@@ -65,7 +76,48 @@ def read(
...
+
+        - A sequence of coordinates is accepted, one per indexed dimension.
+        - The sequence length must be less than or equal to the number of
+          indexed dimensions.
+        - If the sequence is shorter than the number of indexed coordinates,
+          then no constraint (i.e. ``None``) is used for the remaining
+          indexed dimensions.
+        - Specifying ``None`` or an empty sequence (e.g. ``()``) represents
+          no constraints over any dimension, returning the entire dataset.
+          TODO: https://github.com/single-cell-data/TileDB-SOMA/pull/910
```
johnkerl, line 103: this issue is now closed -- ?

python-spec/src/somacore/data.py
```diff
@@ -65,7 +76,48 @@ def read(
...
+
+        Each dimension may be indexed as follows:
+
+        - ``None`` or ``slice(None)`` places no constraint on the dimension.
+        - Coordinates can be specified as a scalar value, a Python sequence
+          (``list``, ``tuple``, etc.), a ``ndarray``, an Arrow array, and
```
johnkerl, line 109: a `ndarray` -> an `ndarray` (the word is prounced _enn dee uh ray_ which begins with a vowel sound, so euphony requires _an_)

python-spec/src/somacore/data.py
```diff
@@ -65,7 +76,48 @@ def read(
...
+          similar objects (as defined by ``SparseDFCoords``).
+        - Slices are doubly inclusive: ``slice(2, 4)`` means ``[2, 3, 4]``,
+          not ``[2, 3]``.  Slice *steps* may not be used: ``slice(10, 20, 2)``
+          is invalid.  ``slice(None)`` places no constraint on the dimension.
+          Half-specified slices like ``slice(None, 99)`` and ``slice(5, None)``
+          specify all indices up to and including that value, and all indices
+          starting from and including the value.
+        - Negative indexing is not supported.
+          TODO: What if the domain includes negative numbers?
```
johnkerl, line 118: we forbid that

https://github.com/single-cell-data/TileDB-SOMA/issues/519

and if that isn't explicit here in the spec it should be

also -- discussion at the time with @atolopko-czi @aaronwolen @bkmartinjr and perhaps others:

* initially we wanted `soma_joinid` to be `uint64` so negative would be impossible
* that works in Python but for R we couldn't do unsigned so we went with `int64` with forbidding the negative side

I'm happy to take this as an action item on a separate PR to be clearer about forbidden negative indices in this spec repo -- just lmk if you don't want to do that on this PR (which is fine with me) & i'll open a separate PR

python-spec/src/somacore/data.py
```diff
@@ -65,7 +76,48 @@ def read(
...
+          Negative values are treated as ordinary indices, e.g. ``slice(-1, 5)``
```
johnkerl, line 119: no

python-spec/src/somacore/data.py
```diff
@@ -158,9 +224,39 @@ def read(
...
+        coordinates must be less than or equal to the number of dimensions.
+        For example, if the array is 10×20, acceptable values of ``coords``
+        include ``()``, ``(3, 4)``, ``[slice(5, 10)]``, and
+        ``[slice(5, 10), slice(6, 12)]``.  Slice indices are doubly-inclusive.
+
+        :param coords: A per-dimension sequence of coordinates defining
+            the range to read.
+        :param batch_size: The size of batches that should be returned
+            from a read. See :class:`options.BatchSize` for details.
+            XXX How does this work if this returns a single Tensor?
```
johnkerl, line 239: please spell `XXX` as `TODO` to make it a more consistently found `grep` target

python-spec/src/somacore/data.py
```diff
@@ -158,9 +224,39 @@ def read(
...
+        :param partitions: Specifies that this is part of a partitioned read,
+            and which partition to include, if present.
+        :param result_order: The order to return the results in.
+
+        **Indexing:**
+
+        Indexing is performed on a per-dimension basis.
+
+        - A sequence of coordinates is accepted, one per dimension.
+        - The sequence length must be less than the number of dimensions.
```
johnkerl, line 249: less than _or equal to_

python-spec/src/somacore/data.py
```diff
@@ -158,9 +224,39 @@ def read(
...
+        - If the sequence is shorter than the number of dimensions, the
+          remaining dimensions are unconstrained. (Thus, if an empty sequence
+          is provided, the entire array will be returned.)
+
+        Each dimension may be indexed by value or slice:
+
+        - Slices are doubly-inclusive.
```
johnkerl, line 256: doubly-inclusive -> doubly inclusive

python-spec/src/somacore/data.py
```diff
@@ -158,9 +224,39 @@ def read(
...
+        - Half-specified slices include all data up to or starting from the
+          specified bound, inclusive.
+        - Negative indexing is unsupported.
```
johnkerl, line 259: correct

python-spec/src/somacore/data.py
```diff
@@ -205,7 +307,7 @@ def read(
         result_order: options.ResultOrderStr = _RO_AUTO,
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> "SparseRead":
-        """Reads a subset of the object in one or more batches.
+        """Reads a subset this DenseNDArray in batches.
```
johnkerl, line 310: Reads a subset _of_ ...

python-spec/src/somacore/data.py
```diff
@@ -215,7 +317,37 @@ def read(
...
             # -> a single flattened sparse CSR matrix
 
-        TODO: Additional per-param documentation.
+        :param coords: A per-dimension sequence of coordinates defining
+            the range to be read.
+        :param batch_size: The size of batches that should be returned
+            from a read. See :class:`options.BatchSize` for details.
+        :param partitions: Specifies that this is part of a partitioned read,
+            and which partition to include, if present.
+        :param result_order: The order to return the results in.
```
johnkerl, line 326: give example values; mysterious otherwise

python-spec/src/somacore/data.py
```diff
@@ -215,7 +317,37 @@ def read(
...
+
+        **Indexing:**
+
+        Indexing is performed on a per-dimension basis.
+
+        - A sequence of coordinates is accepted, one per dimension.
+        - The sequence length must be less than the number of dimensions.
```
johnkerl, line 333: less than _or equal to_

johnkerl (changes requested):

python-spec/src/somacore/data.py
johnkerl: Actually I went ahead and created #121 -- if you do that work on this PR, fanastic & you can close the issue -- otherwise we can do a follow-on PR.

python-spec/src/somacore/data.py
thetorpedodog: Redid all of these to just say "the specified subarray" since you already know what type you're reading.

python-spec/src/somacore/data.py
thetorpedodog: Linked to `ResultOrder`

python-spec/src/somacore/data.py
```diff
@@ -153,14 +215,41 @@ def read(
         self,
         coords: options.DenseNDCoords,
```
thetorpedodog, line 216: Is there a reason that we specify a default for `SparseNDArray.read` but not `DenseNDArray.read`?

python-spec/src/somacore/data.py
```diff
@@ -46,15 +46,26 @@ def create(
...
+        """
         raise NotImplementedError()
 
     # Data operations
 
     @abc.abstractmethod
     def read(
         self,
-        coords: Optional[options.SparseDFCoords] = None,
+        coords: Optional[options.SparseDFCoords] = (),
```
thetorpedodog, line 68: Since we now have the default of `()` that unambiguously means "everything", does this need to be `Optional` (i.e., no longer accept `None` as a value)? Consider: `DenseNDCoords` is currently *not* marked as `Optional`.

johnkerl: as long as it's optional put-nothing-at-all (which it is) that's is fine with me

python-spec/src/somacore/data.py
johnkerl: @thetorpedodog we are on the same page

* didn't have a specific example before
* it should
* but value-filter language is an impl-repo thing and this is a spec-repo PR so it's not ok for us to put _any_ examlpe here

which is unfortunate

all i'm doing is pointing that out

i didn't ask any action from you here & still am not -- just pointing this out as food for thought

the real solution for the "how do i spell a value filter in a query" is going to have to be unanswerable in a docstring (given that this query _code_ is in the _spec_ repo)

the impl stuff will belong in impl-repo examples

johnkerl: and ok i guess i _do_ have an ask after all:

let's go ahead say something of the form "value-filter syntax is implementation-dependent -- please see implementation docs for examples"

python-spec/src/somacore/data.py
johnkerl: no argument; i get the idea we're all doing parallel development

as long as the TODO is gone that's all i ask

python-spec/src/somacore/data.py
johnkerl: @thetorpedodog i saw it only b/c it was flagged in green by the GitHub PR tool

anyone coming along later won't have that benefit

my ask is that we not mix `XXX` and `TODO` when we could just spell `TODO`

this is an ask of just a few keystrokes

python-spec/src/somacore/data.py
johnkerl: my guess is that was inadvertent

@bkmartinjr to verify

python-spec/src/somacore/data.py
thetorpedodog: Correct:

✅ `some_df.read()`
✅ `some_df.read(coords=())`
❌ `some_df.read(coords=None)`

I’ve put this change in for now as its own commit so if we don’t want it, it’s easy to back out.

python-spec/src/somacore/data.py
thetorpedodog: Added the default (in the same separate commit).

python-spec/src/somacore/data.py
thetorpedodog: Added. If we *do* want a canonical value-filter syntax then this can be updated.

johnkerl (approved):

python-spec/src/somacore/data.py
thetorpedodog: @bkmartinjr (and any other readers): another thing to check in on (goes with https://github.com/single-cell-data/SOMA/pull/119#discussion_r1103107462)

python-spec/src/somacore/data.py
thetorpedodog: Given the [discussion on bug 121](https://github.com/single-cell-data/SOMA/issues/121#issuecomment-1426485914), I think this probably warrants reinstating language around negative indexing:

> Negative numbers as indices or parts of slices are used as raw domain values and are not treated specially, unlike traditional Python sequence indexing. For instance, `slice(-10, 3)` indicates the range from −10 to 3 on the given dimension.

This particular bit of language only belongs in DataFrame, since in the NDArrays, negative indices do not appear at all. (And if somebody tries to index their DataFrame’s `soma_join_id` by `[-10:3]`, then that’s fine; they’ll get indices 0–3 inclusive as if they did so on any other non-negative dimension.)

python-spec/src/somacore/data.py
bkmartinjr: I am _very_ good with this change.

python-spec/src/somacore/base.py
```diff
@@ -61,6 +68,18 @@ def metadata(self) -> MutableMapping[str, Any]:
         """
         raise NotImplementedError()
 
+    @property
+    @abc.abstractmethod
+    def mode(self) -> options.OpenMode:
+        """Returns the mode this object was opened in."""
```
bkmartinjr, line 74: do you want to expand to match the docstring in `tiledbsoma`, which inclues "... either 'r' or 'w'."

python-spec/src/somacore/data.py
```diff
@@ -46,15 +46,26 @@ def create(
         platform_config: Optional[options.PlatformConfig] = None,
         context: Optional[Any] = None,
     ) -> Self:
-        """Creates a new DataFrame."""
+        """Creates a new ``DataFrame`` at the given URI.
```
bkmartinjr, line 49: do we want to include info on the requirements & defaulting around soma_joinid?

i.e.,
* soma_joinid must have type pyarrow.int64
* if soma_joinid is omitted from the user-provided schema, it will be added
* it may be an indexed column

python-spec/src/somacore/data.py
```diff
@@ -65,7 +76,46 @@ def read(
...
+        - Specifying an empty sequence (e.g. ``()``, the default) represents
+          no constraints over any dimension, returning the entire dataset.
+
+        Each dimension may be indexed as follows:
+
+        - ``None`` or ``slice(None)`` places no constraint on the dimension.
+        - Coordinates can be specified as a scalar value, a Python sequence
+          (``list``, ``tuple``, etc.), a NumPy ndarray, an Arrow array, or
+          similar objects (as defined by ``SparseDFCoords``).
+        - Slices are doubly inclusive: ``slice(2, 4)`` means ``[2, 3, 4]``,
```
bkmartinjr, line 112: `doubly-inclusive` is a term that I do not often see in Python docs - more common is `half-open` or `closed`, to track with traditional interval/range nomenclature.

Suggested change for clarity:   "Slices are doubly inclusive (i.e., specify a closed range): ..."

python-spec/src/somacore/data.py
```diff
@@ -46,15 +46,26 @@ def create(
...
+        :param uri: The URI where the ``DataFrame`` will be created.
+        :param schema: Arrow schema defining the per-column schema. This schema
+            must define all columns, including columns to be named as index
+            columns.  If the schema includes types unsupported by the SOMA
+            implementation, an error will be raised.
+        :param index_column_names: A list of column names to use as user-defined
+            index columns (e.g., ``['cell_type', 'tissue_type']``).
+            All named columns must exist in the schema, and at least one
+            index column name is required.
+        """
```
bkmartinjr, line 60: quite a few methods are missing lifecycle tags

python-spec/src/somacore/data.py
bkmartinjr: ditto for other places we use the term.

bkmartinjr (approved):
Minor comments for sake of improving clarity. No objections to content.  Thanks for doing this!

python-spec/src/somacore/data.py
thetorpedodog: added to all methods

python-spec/src/somacore/data.py
thetorpedodog: Changed to "closed" in all locations.

python-spec/src/somacore/data.py
thetorpedodog: Added this information.

python-spec/src/somacore/data.py
thetorpedodog: update: yes

PR #123, opened 2023-02-13
[python] Enable chaining on certain write methods.

thetorpedodog:
Returns `self` from a few methods to enable chaining:

- DataFrame/DenseNDArray/SparseNDArray.write
- Collection.set

This enables useful chaining:

    some_coll.set('x', ...).set('y', ...)

    my_df = DataFrame.create(...).write(...)
    do_stuff_with(my_df)

---

I remember this being something we discussed a while back that got lost in other things, so here it is back again.


bkmartinjr (approved):
Very much in favor.  Thanks!

Issue #121, opened 2023-02-10
Clarify/implement forbiddenness of negative `soma_joinid` in the spec


johnkerl:
Tracking issue for https://github.com/single-cell-data/SOMA/pull/119#discussion_r1102882763 in the event that feedback needs addressing in a separate PR.

See also https://github.com/single-cell-data/TileDB-SOMA/issues/519.

-------

Analysis results in support of https://github.com/single-cell-data/SOMA/pull/124 and https://github.com/single-cell-data/TileDB-SOMA/pull/932:

* Spec:
  * The [spec language](https://github.com/single-cell-data/SOMA/blob/python-v0.0.0a13/abstract_specification.md) is almost correct: it says _Numeric index types (eg, offset indexing into dense arrays) are specified with `int64` type and a domain of `[0, 2^63-1)`. In other words, positive `int64` values are used for offset indexing._ All this needs is the typo _positive_ to be changed to _non-negative_: https://github.com/single-cell-data/SOMA/pull/124
  * The spec docstring info could use to make this explicit: also on https://github.com/single-cell-data/SOMA/pull/124
* `TileDB-SOMA` Python impl:
  * Currently allows negative `soma_joinid`, contrary to the spec: fixed by https://github.com/single-cell-data/TileDB-SOMA/pull/932 
* `TileDB-SOMA` R impl:
  * Is already correct:
    * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/apis/r/R/SOMADataFrame.R#L56-L64
    * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/apis/r/R/utils-arrow.R#L130-L142
  * Further proof it's already correct:
    * https://gist.github.com/johnkerl/547d13fb3782fa71cf50c37136e585d9
    * Copied from https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/apis/r/tests/testthat/test_SOMADataFrame.R#L2
    * Array-schema output added as a comment on that gist reveals domain was written in conformance with the spec: `Dim(name='soma_joinid', domain=(0, 9223372036854775807), ...)`

thetorpedodog:
For future consideration:

Since we allow floating-point and string dimensions, it would make sense to me that we should allow users to store data with a domain that goes into the negative numbers in a DataFrame.

[...]

PR #124, opened 2023-02-14
Clarify non-negative `soma_joinid` in docstring and spec

johnkerl:
Described fully at https://github.com/single-cell-data/SOMA/issues/121

### DIFF
--- a/python-spec/src/somacore/data.py
+++ b/python-spec/src/somacore/data.py
@@ -52,10 +52,10 @@ class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
         [lifecycle: experimental]
 
         The schema of the created dataframe will include a column named
-        ``soma_joinid`` of type ``pyarrow.int64``, with negative values disallowed.
-        If a ``soma_joinid`` column is present in the provided schema, it must be of
-        the correct type.  If no ``soma_joinid`` column is provided, one will be
-        added. This column may be used as an indexed column.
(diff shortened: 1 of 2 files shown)


bkmartinjr (approved):

python-spec/src/somacore/data.py
```diff
@@ -52,10 +52,10 @@ def create(
...
 
         The schema of the created dataframe will include a column named
-        ``soma_joinid`` of type ``pyarrow.int64``. If a ``soma_joinid`` column
-        is present in the provided schema, it must be of the correct type. If no
-        ``soma_joinid`` column is provided, one will be added. This column may
-        be used as an indexed column.
+        ``soma_joinid`` of type ``pyarrow.int64``, with negative values disallowed.
+        If a ``soma_joinid`` column is present in the provided schema, it must be of
+        the correct type.  If no ``soma_joinid`` column is provided, one will be
+        added. This column may be used as an indexed column.
```
thetorpedodog, line 58: just a reflow with a changed word or two:

```suggestion
        ``soma_joinid`` of type ``pyarrow.int64``, with negative values
        disallowed.  If a ``soma_joinid`` column is present in the provided
        schema, it must be of the correct type.  If no ``soma_joinid`` column
        is provided, one will be added.  It may be used as an indexed column.
```

thetorpedodog (approved):

PR #128, opened 2023-02-15
[python] Make `Slice` protocol not `runtime_checkable`

thetorpedodog:
Since `range` has the members `start`/`stop`/`step`, that meant that `isinstance(range(x), Slice)` would return True, when a `range` is *not* a `Slice`.  Using `abc` and manually registering `slice` fixes this.

This also unrestricts `Slice`s from being `Comparable`, since the built-in `slice` type does not have this restriction, and adds the `is_slice_of` function, taken from `tiledbsoma`.

### DIFF
--- a/python-spec/src/somacore/types.py
+++ b/python-spec/src/somacore/types.py
@@ -1,7 +1,8 @@
 """Type and interface declarations that are not specific to options."""
 
-from typing import Any, Optional, TypeVar, Sequence
-from typing_extensions import Protocol, Self, runtime_checkable, TypeGuard
+import sys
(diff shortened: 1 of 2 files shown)


thetorpedodog:
Update: Since [mypy doesn’t currently understand `SomeABC.register(other_type)`](https://github.com/python/mypy/issues/2922), that approach Did Not Work well in reality. Instead, I’ve just removed the `runtime_checkable`ility of the protocol and made `is_slice_of` the canonical type-checking method.

johnkerl (approved):

thetorpedodog:
Oops, had not noticed the approval. Reverting to the approved state (i.e. no longer removing `Comparable`) and saving that for another change.

PR #136, opened 2023-02-21
Make `()` the default value for `AxisQuery.coords`.

thetorpedodog:
Since `coords` values are no longer `Optional` (and the `(slice(None),)` workaround is no longer needed), this updates the declared API for AxisQuery to also not be `Optional` (i.e., to request everything, the `()` should be used).


atolopko-czi (approved):
LGTM

bkmartinjr (approved):

PR #138, opened 2023-02-22
[python] Add support for floats in axis query

thetorpedodog:
As of recently [1], TileDB-SOMA now supports floats when indexing dataframes.  We should no longer forbid them in axis queries.

[1]: https://github.com/single-cell-data/TileDB-SOMA/pull/968

Context: issue https://github.com/single-cell-data/TileDB-SOMA/issues/960


bkmartinjr (approved):
Yay!

johnkerl (approved):
🚢

PR #141, opened 2023-02-24
[python] Add `bool` to supported types and improve type specification

thetorpedodog:
This adds boolean to the supported sparse coordinate types and also improves the type definition of coordinate types to ensure that the values in each type are homogeneous:

    BadCoord = Union[int, str]
    BadCoords = Union[BadCoord, Sequence[BadCoord]]
    # ["this", 1] is bad, but is a valid as a BadCoords value

    GoodCoord = TypeVar("GoodCoord", int, str)
    GoodCoords = Union[GoodCoord, Sequence[GoodCoord]]
    # ["this", 1] is bad, and is *not* valid as a GoodCoords value
    # ["this", "that"] is a valid as a GoodCoords value
    # [1, 2] is valid as a GoodCoords value

---

I don’t see a bug entry for this but it was something we discussed earlier today, and in the process I realized another thing we could improve.

Context: #960

### DIFF
--- a/python-spec/src/somacore/options.py
+++ b/python-spec/src/somacore/options.py
@@ -5,7 +5,7 @@ SOMA types that require them, not reimplemented by the implementing package.
 """
 
 import enum
-from typing import Any, Mapping, Optional, Sequence, Union
+from typing import Any, Mapping, Optional, Sequence, TypeVar, Union
 
 import attrs
 import numpy as np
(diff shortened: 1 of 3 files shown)


python-spec/src/somacore/options.py
```diff
@@ -110,36 +110,42 @@ class ResultOrder(enum.Enum):
...
-    types.Slice[str],
-    types.Slice[bytes],
-    types.Slice[np.datetime64],
-    types.Slice[pa.TimestampType],
+    SparseCoord,
+    Sequence[SparseCoord],
+    types.Slice[SparseCoord],
     pa.Array,
     pa.ChunkedArray,
     npt.NDArray[np.integer],
```
bkmartinjr, line 151: should we also have a numpy-array-like for `np.float`, `np.datetime64`, etc?

bkmartinjr (approved):
Changes look good.  Led to a question about possibly missing point indexing types to match the new slices (eg, if we have slice(float), why not np.array(float)?)

python-spec/src/somacore/options.py
thetorpedodog: This is a good idea but I think it makes sense for a future change. (Also, I am not familiar enough with the numpy typing system at the moment to do the right thing here.)

johnkerl: https://github.com/single-cell-data/SOMA/issues/142

johnkerl (approved):
🚢

PR #143, opened 2023-02-28
[python] Explicit shaping for `DataFrame`

johnkerl:
Context: https://github.com/single-cell-data/TileDB-SOMA/issues/961

This is a prerequisite for https://github.com/single-cell-data/TileDB-SOMA/pull/1027.

Note that for `SparseNDArray` it sufficed to have a slotwise int `shape`: indexing is only by `int64` and we took the min to be 0, leaving only the max to be specified. Here, for `DataFrame`, indices can be of various types so we truly must do parameterizations in terms of domains -- min-max pairs on each dim slot.

### DIFF
--- a/python-spec/src/somacore/data.py
+++ b/python-spec/src/somacore/data.py
@@ -70,13 +70,17 @@ class DataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
             All named columns must exist in the schema, and at least one
             index column name is required.
 
-        :param domains: An optional sequence of min-max tuples specifying the
-            domain of each index column, including room for any intended future
-            appends.  If provided, this sequence must have the same length as
-            `index_column_names`, and the index-column domains will be as specified.
-            If omitted entirely, or if ``None`` in a given dimension, the
-            corresponding index-column domain will use the maximum possible values.
-            This makes a `SOMADataFrame` growable.
+        :param domains: An optional sequence of tuples specifying the domain of each
+            index column. Each tuple should be a pair consisting of the minimum and
+            maximum values storable in the index column. For example, if there is a
+            single int64-valued index column, then ``domains`` might be ``[(100,
+            200)]`` to indicate that values between 100 and 200, inclusive, can be
+            stored in that column.  If provided, this sequence must have the same
+            length as `index_column_names`, and the index-column domains will be as
+            specified.  If omitted entirely, or if ``None`` in a given dimension,
+            the corresponding index-column domain will use the minimum and maximum
+            possible values for the column's datatype.  This makes a
+            ``SOMADataFrame`` growable.
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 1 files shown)


python-spec/src/somacore/data.py
```diff
@@ -58,14 +59,24 @@ def create(
...
             must define all columns, including columns to be named as index
             columns.  If the schema includes types unsupported by the SOMA
             implementation, an error will be raised.
+
         :param index_column_names: A list of column names to use as user-defined
             index columns (e.g., ``['cell_type', 'tissue_type']``).
             All named columns must exist in the schema, and at least one
             index column name is required.
+
+        :param domains: An optional sequence of min-max tuples specifying the
```
bkmartinjr, line 73: same comments as on the `tiledbsoma` PR review - the param description is confusing.  Suggest making it consistent with the other PR

python-spec/src/somacore/data.py
```diff
@@ -45,6 +45,7 @@ def create(
         *,
         schema: pa.Schema,
         index_column_names: Sequence[str] = (options.SOMA_JOINID,),
+        domains: Optional[Sequence[Optional[Tuple[Any, Any]]]] = None,
```
bkmartinjr, line 48: the thing missing from this is access to the domain/extent on an existing DataFrame.

We used to have `shape` (it still exists on NDArray), and I thought we intended to bring that back once the DataFrame had added this first-class concept.

So, where is the `shape` property equivalent?   And what should it be called?   (`domain_extent`?)

CC: @thetorpedodog

bkmartinjr (dismissed):
the major open issue IMHO is the addition of the domain, without a property to access it on existing DataFrames.  We need the accessor (much like NDArray has `shape` on both `create` and as a read-only property on existing NDArrays).

python-spec/src/somacore/data.py
```diff
@@ -170,6 +185,14 @@ def index_column_names(self) -> Tuple[str, ...]:
         """
         raise NotImplementedError()
 
+    @property
+    def domains(self) -> Tuple[Tuple[Any, Any], ...]:
```
johnkerl, line 189: I'm now thinking this should be singular `domain` in line with tiledb usage.

If everyone's agreed I'll of course make the same change in the impl PR https://github.com/single-cell-data/TileDB-SOMA/pull/1027.

python-spec/src/somacore/data.py
```diff
@@ -170,6 +185,14 @@ def index_column_names(self) -> Tuple[str, ...]:
         """
         raise NotImplementedError()
 
+    @property
```
thetorpedodog, line 188: don’t forget to add an `@abc.abstractmethod` here.

thetorpedodog (commented):
I have no objections here (apart from the one note below); if Brian is happy with it I am.

python-spec/src/somacore/data.py
bkmartinjr: I prefer the singular as well.  @thetorpedodog ?

bkmartinjr:
> if Brian is happy with it I am

I don't think Brian is on the review - add him?

(joking - I assume this was a typo)

bkmartinjr (approved):

PR #149, opened 2023-03-03
Change collection child type to "kind" name.

thetorpedodog:
Companion to https://github.com/single-cell-data/TileDB-SOMA/pull/1045.


atolopko-czi (approved):
Ah, thanks for catching this additional update!

PR #151, opened 2023-03-06
[python] Convert docstrings to Google style

thetorpedodog:
This also expands the docstrings for Experiment and Measurement.

---

In this change, I also changed the format of maturity levels to make them look like just another docs stanza. If we don’t like it, it can be reverted to the `[lifecycle: whatever]` style.

Part of https://github.com/single-cell-data/TileDB-SOMA/issues/974.

### DIFF
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -131,6 +131,8 @@ class SOMAObject(metaclass=abc.ABCMeta):
         This is also called automatically by the Python interpreter via
         ``__del__`` when this object is garbage collected, so the implementation
         must be idempotent.
+
+        Lifecycle: experimental
         """
         # Default implementation does nothing.
 
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -28,13 +28,15 @@ class SOMAObject(metaclass=abc.ABCMeta):
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> Self:
         """Opens the SOMA object of this type at the given URI.
-        [lifecycle: experimental]
 
-        :param uri: The URI of the object to open.
-        :param mode: The mode to open this in, either `r` or `w`.
-        :param context: The Context value to use when opening the object.
-        :param platform_config: Platform configuration options specific to
-            this open operation.
+        Args:
+            uri: The URI of the object to open.
+            mode: The mode to open this in, either `r` or `w`.
+            context: The Context value to use when opening the object.
+            platform_config: Platform configuration options specific to
+                this open operation.
+        Returns: The SOMA object, opened for reading.
+        Lifecycle: experimental
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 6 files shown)


bkmartinjr:
Question:  is the `[lifecycle: whatever]` supposed to be machine parsable?  I.e., does the format matter, or just the human-readable content?

python-spec/src/somacore/base.py
```diff
@@ -111,9 +124,13 @@ def closed(self) -> bool:
 
     def close(self) -> None:
         """Releases any external resources held by this object.
-        [lifecycle: experimental]
 
-        An implementation of close must be idempotent.
+        For objects opened for write, this also finalizes the write operation
```
bkmartinjr, line 128: missing lifecycle stanza?

python-spec/src/somacore/collection.py
```diff
@@ -83,43 +87,55 @@ def add_new_collection(
...
+
+            # flat_ns_coll URI is "flat://authority/key" in a flat namespace
+            # where relative paths are unsupported.
+            flat_ns_coll.add_new_collection("flat child")
+            # The URI of the child collection might be:
+            # "flat://authority/flat_child"
+
+        The way the URI is constructed is left unspecified so that an
+        implementation can create a URI based on its own needs. Users should
+        directly get the URI of the new child if they need it; they should never
```
bkmartinjr, line 122: worth providing an example of how to get the URI?

thetorpedodog:
With the disclaimer that this is just my opinion: I expect it to be only human-readable. If we want to have a machine-parseable version of the lifecycle (which seems like a niche use case), I would have it in a function attribute (maybe set `.soma_lifecycle` in a decorator) rather than the docstring, something like:

```python
from . import lifecycle

@lifecycle.experimental
def some_func(...):
  ...

some_func.soma_lifecycle
# -> experimental
```

python-spec/src/somacore/data.py
```diff
@@ -323,17 +360,18 @@ def write(
...
-        :param values: The values to be written to the subarray.  Must have
-            the same shape as ``coords``, and matching type to the array.
-        :return: ``self``, to enable method chaining.
+        Args:
+            coords: A per-dimension tuple of scalars or slices
+                defining the bounds of the subarray to be written.
+                See :meth:`read` for details about indexing.
+            values: The values to be written to the subarray.  Must have
+                the same shape as ``coords``, and matching type to the array.
+        Returns: ``self``, to enable method chaining.
```
bkmartinjr, line 373: Some of the docstrings have a blank line between stanzas, some do not.  Do we need to be consistent?

I have a _very_ slight bias towards no blank line, to condense information when viewing in an IDE.

bkmartinjr:
> I expect it to be only human-readable.

Given this, I also prefer your proposed style (i.e., a new stanza).

python-spec/src/somacore/data.py
thetorpedodog: I was doing this on a not-entirely-systematic method of “does it seem like there’s a lot of text here” as to whether to include or omit the extra line (so if there are really long segments, or ones that use multiple paragraphs, include a blank line, but otherwise not). There are situations like `SparseNDArray.write` where it is almost necessary given the contents.

python-spec/src/somacore/base.py
thetorpedodog: Added.

python-spec/src/somacore/collection.py
thetorpedodog: Added.

johnkerl:
I completely agree re human-readable, and that if we want programmatic access we could do a `soma_lifecycle()` or somesuch,

johnkerl (approved):

python-spec/src/somacore/data.py
bkmartinjr: OK, works for me.

bkmartinjr (approved):

PR #153, opened 2023-03-22
[python] Add more lifecycle-experimental indicators to docstrings

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/issues/1088

See also https://github.com/single-cell-data/TileDB-SOMA/pull/1153


johnkerl:
Duplidate of #154

PR #161, opened 2023-05-08
Add initial directory and README for SOMA RFCs/design docs.

maniarathi:
This directory can be used to capture design documents and proposals about the SOMA spec and can be used as a source of truth to revisit technical decisions in a central location.


Shelnutt2 (approved):
lgtm :tada:

PR #162, opened 2023-05-09
SOMA Profiler RFCs

beroy:


bkmartinjr:
@beroy - one thing that would help me review this is some additional context and background information on the requirements and definition of success.  Questions top of mind:

1. Is the primary use case automating a smoke test to detect regressions as part of CI?  Or is the goal to help diagnose/debug regressions once they are detected?  Or both?
2. If your goal is a diagnostic tool:  my experience is that regressions are often I/O or schema related, or are highly obscured by concurrency issues.  I am wondering if a focus on CPU/memory profiling is going to be useful in practice (vs. for example, the built-in TileDB query statistics). Put another way, if your goal is a diagnostic tool, is this the biggest bang for the buck?   (very curious to hear what @Shelnutt2 and @gspowley think as I'm sure they have internal tooling that has covered some of this ground).
3. If your goal is detection/smoke testing for regressions, and you think a historical log of past runs will be useful, do you want to capture core DB statistics?

Lastly (minor point) - we should probably discuss if this belongs in the TileDB-SOMA repo, or as a peer repo (motivated by a goal of keeping each repo simpler/focused - i.e., the old monorepo debate :-)

beroy:
@bkmartinjr, thanks a lot for the great comments. We had a design review last week and discussed some of the related issues but did not cover all the questions here. I try to address them here:
1, 2) The main goals of this are a) figuring the hot spots (perf/mem) in our code stack (either python or R) and try to reduce the bottleneck b) the ability of detect regressions as a part of CI. The generic profiler only allows detection in (b) and not necessarily diagnosis as you mentioned. Diagnosis could be a lot more complicated than what software stack frame graphs can provide.
3) That's a great point. I did not think about it but thinking more, I really think that can be very useful if I can do that as custom profiler! That means we still have the general architecture but we also have a TileDB profiler that gets attached to the app and collects the TileDB stats. Not sure how possible it is to access TileDB API from outside a process. In the worst case we can add an access point or just a flag for doing so.

Finally, this repo suggestion seems to come from an agreement between TileDB and CZI for sharing RFCs. @maniarathi can comment here.

bkmartinjr:
Hi @beroy,

> design review last week

Did this review include the TileDB team?  Most of the code is in their DB stack (not in SOMA or Census code), and that is also where most of the complexity and performance sensitive code is (e.g, github.com/TileDB/TileDB and github.com/TileDB/TileDB-Py).  The core TileDB team has extensive experience with performance work, and I presume have a tool stack already in place. I just want to make sure we are filling a gap that exists, and that we benefit from the existing knowledge and tools to build upon.

> TileDB stats

AFAIK, you can't capture them from outside the process, but enabling in-process collection has very little overhead and they are designed to be captured in the normal course of using the DB.  Given that you must run the system with your own driver process, I suggest just capturing them as part of that driver process.  The total size of stats summary is tiny and can be written to logs as part of running the core tests.

There are some near-term caveats to this due to our double-instantiation of the core DB, @gspowley and others can talk you through their plans to centralize everything (tactically,just capture stats fro both instances, or perhaps even ignore the second if you are focused on read-only perf).  The stats API in SOMA is already exposed via `tiledbsoma.tiledbsoma_stats_*`, and there are equivalent API in the `tiledb` package for the second DB instance.

johnkerl:
> @gspowley and others can talk you through their plans to centralize everything

@bkmartinjr the current POC is @nguyenv

beroy:
@bkmartinjr. TileDB people were not in the review but I discuss their profiling tools with @gspowley. While their tool capture some of the needed stuff the main focus is on distributed nodes. Also my goal is to capture the breakdown across python/R and C++. Your suggestion about collecting DB stats makes total sense (directly log them from the driver). BTW, here by core DB do you mean TileDB? Also, I'm not familiar with the double-instantiation plan. Will check with them

thetorpedodog:
two non–content-related notes from me (these apply to both open RFCs right now so I am pasting this note in both places):

1. Rebase so that you get the pre-commit format verification thingy
2. (Optional) Consider formatting to one sentence per line. I did this in 3b5891c678e9d08f4c8817cb6426f82c02675cf0, but I didn’t formally write it down, so it’s not a hard Rule but it is a format I have found useful for editing.

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+**FlameGraph:** A flame graph visualizes a distributed request trace and represents each service call that occurred during the request’s execution path with a timed, color-coded, horizontal bar.
+
+# Requirements
+
+**Basic requirements:**
+
+- Our solution should be able to provide timing metrics for a full run.
+- Our solution should be able to provide memory metrics for a full run.
+- Our solution should be able to be incorporated into a release pipeline to protect against / debug regressions in code
```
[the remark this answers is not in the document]
atolopko-czi: capturing env/context for the run should be a requirement (host info, etc.)

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+## Engineering Steps
+
+- We need to make sure all nodes across our controlled environment for profilers have **_time_** utility. Also need to figure out the right flags for them to extract the memory in addition to performance metrics.
+- A way to have a central DB (Filesystem only or **_SQLiteDict_**) across the environments (EC2 and developer machines).
+- As more custom profilers are added to the system, they must be tested and added to the controlled environment gradually.
+
+## Deployment Steps
+
+- **Internal performance tracking**
+  - We start the generic profiler on a small controlled environment (like github tests and regular performance runs before the releases) a shared DB
```
atolopko-czi, line 133: "a shared DB"... incomplete?

atolopko-czi (changes requested):
A few requests for clarifications and decisions about scope of the RFC.

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+A metric for performance and memory regression should be designed and implemented. The use of custom profilers can be conditional on detection of regression. The idea being once a regression is detected, we rerun the process with custom profilers to avoid slowing down normal execution.
+
+## Security & Data Privacy
+
+The central DB can pose a risk if that gets compromised as it reveals the history of all of the Single Cell Dataset runs.
+
+## Estimated Cost
+
+This design does not require adding new machines (nodes). It only relies on installation of the **_time_** utility and custom profilers across the nodes. The other cost is the storage needed for the central DB.
```
ebezzi, line 120: Even if it doesn't belong to the `cost` section, we need to figure out the ownership of the centralized database. As far as I know, for the SOMA project, we opted to keep most of the assets in GitHub so that we don't tie them to e.g. a specific AWS account. This would be difficult if we decided to use a "real" database like DynamoDB.

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+To address the mentioned basic requirements, at the top level, we use a generic and simple profiler which only tracks end-to-end execution time and memory metrics (peak memory, page faults, etc.) of the executed queries. This profiler uses a commonly available tool such as **_time_** to keep track of this high level information across days and releases in a small database (the early version can be using just the filesystem (key will be the filename and value will be the file content) and for then next phase, we can use [SqliteDict](https://pythonhosted.org/sqlite_object/sqlite_dict.html) K/V database, TileDB or DynamoDB as other alternatives) across different platforms. As the profiler runs the application process, a new record will be generated and stored. The record associated with each run has the following schema:
+
+- **Key:**
+  - _Date_
+  - _Time_
+  - _Process Information (application and its arguments)_
+- **Value:**
+  - _Host information_
+  - _User Defined MD_
```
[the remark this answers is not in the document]
beroy: Meta data

rfcs/profiler.md
beroy: You are right. The final decision was to keep this on github. I'll update it

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+# Requirements
+
+**Basic requirements:**
+
+- Our solution should be able to provide timing metrics for a full run.
+- Our solution should be able to provide memory metrics for a full run.
+- Our solution should be able to be incorporated into a release pipeline to protect against / debug regressions in code
+
+**Advanced requirements:**
```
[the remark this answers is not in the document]
beroy: It's added to the Generic profile **Value** part

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
+# Tech Spec: Performance Evaluation & Monitoring for Single Cell
+
+# Goal
+
+Single Cell datasets and queries run across multiple platforms (Mac OSes, Linux, AWS VMs, etc.) and across multiple layers of software stack (python, C++ and R as of now!). Performance and memory monitoring for such applications is essential because:
```
johnkerl, line 5:
```suggestion
Single-cell datasets and queries run across multiple platforms (Mac OSes, Linux, AWS VMs, etc.) and across multiple layers of software stack (Python, C++ and R as of now!). Performance and memory monitoring for such applications is essential because all of the following are important:
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+- Detection of regression as new features are added is very important
```
johnkerl, line 7:
```suggestion
- Detection of regression as new features are added
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+- Performance comparison across different platforms and languages to spot memory and execution time hot spots for optimization is critical
```
johnkerl, line 8:
```suggestion
- Performance comparison across different platforms and languages to spot memory and execution time hot spots for optimization
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+- Measuring the system scalability as the workload grows in size is very important
```
johnkerl, line 9:
```suggestion
- Measuring the system scalability as the workload grows in size
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+- Profiling a _suite_ of various operations, which can be run as a unit, reproducibly.
```
johnkerl, line 10:
```suggestion
- Profiling a _suite_ of various operations, which can be run as a unit, reproducibly
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+- ~~Having a tool for customers to monitor and debug their workloads~~
+
+The goal of this process is to provide a multi-layer profiler (consisting of a generic top profiler and multiple custom , language dependent, profilers) to help Single Cell detect potential regression caused by bugs or new releases ~~and also help the customers detect performance issues in their queries~~.
```
johnkerl, line 13:
```suggestion
The goal of this process is to provide a multi-layer profiler -- consisting of a generic top profiler and multiple custom , language-dependent profilers -- to help detect potential regression caused by bugs or new releases of TileDB-SOMA ~~and also help the customers detect performance issues in their queries~~.
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+**Advanced requirements:**
+
+- Our solution should be able to provide a breakdown of how time was spent in different components / queries
+- Our solution should be able to provide a breakdown of how memory was used in different components / queries
+- Our solution should be able to aggregate metrics from multiple components into a roll-up view for internal use (may be useful for the third party as well)
+- ~~[P2] Our solution should be able to be released to clients to help them identify and improve poor performing queries~~
+
+## Future Work
+
+This work is open ended as there is a chance to add more and more custom profilers to the system. Also right now, we plan to use flamegraph to connect custom profilers and the generic main profiler. This can also be extended by using different intermediate formats or objects.
```
johnkerl, line 38:
```suggestion
This work is open-ended as there is a chance to add more and more custom profilers to the system. Also right now, we plan to use flamegraph to connect custom profilers and the generic main profiler. This can also be extended by using different intermediate formats or objects.
```

rfcs/profiler.md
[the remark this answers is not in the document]
johnkerl: metadata

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+    - _Git commit SHA1_
+    - _Description_
+    - _TileDB stats_
+  - _Tag_
+  - _Time (real time, sys time, user time)_
+  - _Memory (peak, max resident, page faults, …)_
+  - _Custom Profiler 1 output_
+  - _Custom Profiler 2 output_
+
+As can be seen, the key is a combination of **_Date_**, **_Time_** and **_Process Info_**. The idea is to make it easy to query across time for a given process. The value consists of executed process detail info, library versions (releases and PR perhaps too). Additionally, the user can add platform information (like Single Cell library versions) and git **_SHA1_** and a _description_ for the benchmark. **_Tag_** is a value linking multiple benchmarks (a suite of runs). Finally, the profile information also includes time and memory metrics. Additionally, it includes the output info of custom profilers. Custom profilers will be discussed in the next section.
```
johnkerl, line 69:
```suggestion
As can be seen, the key is a combination of **_Date_**, **_Time_** and **_Process Info_**. The idea is to make it easy to query across time for a given process. The value consists of executed process detail info, library versions (releases and PR perhaps too). Additionally, the user can add platform information (like SOMA library versions) and git **_SHA1_** and a _description_ for the benchmark. **_Tag_** is a value linking multiple benchmarks (a suite of runs). Finally, the profile information also includes time and memory metrics. Additionally, it includes the output info of custom profilers. Custom profilers will be discussed in the next section.
```

johnkerl: The spec branding is SOMA. The impl branding is TileDB-SOMA. There is no "Single Cell" branding in our terminology.

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+If a more detailed breakdown of the software stack is needed (for example if we suspect a new unexpected hot spot has taken place after a release), we need a tracing profiler to extract time or memory spent/used by each component in our software stack. To achieve this goal, one or more custom profilers (language dependent) can be invoked by the generic profiler. In other words, these profilers work as a plugin into the generic profiler that can be enabled by the user. Once finished, the generic profiler stores the output of these profilers in its database for the users.
+
+### Custom Profiler API
+
+We studied a good number of python profilers including [cProfile](https://docs.python.org/3/library/profile.html), [line_profiler](https://pypi.org/project/line-profiler/), [tracemalloc](https://docs.python.org/3/library/tracemalloc.html), etc. While each of these profilers provides great information, the format and output of them is different and supporting them for languages across different systems can be challenging. Instead, we decided to have custom profilers that use a consistent format for their outputs. This provides a common and useful interface into the generic profiler.
```
johnkerl, line 84:
```suggestion
We studied a good number of Python profilers including [cProfile](https://docs.python.org/3/library/profile.html), [line_profiler](https://pypi.org/project/line-profiler/), [tracemalloc](https://docs.python.org/3/library/tracemalloc.html), etc. While each of these profilers provides great information, the format and output of them is different and supporting them for languages across different systems can be challenging. Instead, we decided to have custom profilers that use a consistent format for their outputs. This provides a common and useful interface into the generic profiler.
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+We decided to use [flamegraph](https://github.com/brendangregg/FlameGraph) as this common interface. The **_framegraphs_** are a very popular interactive way of tracking performance metrics across software components. Therefore, the generic profiler will be given a set of custom profilers (and their arguments) to run and simply expect each profiler to generate a new **_flamegraph_** of the program software stack in a particular location and it adds the generated files to the Database as well. Given the overhead of tracebased profilers, custom profilers are always going to be optional. For example upon detecting a regression in the generic profiler DB, we can rerun the application with the custom profiler to get the **_flamegraphs_** of the application. For R, we can use [xProf](https://github.com/atheriel/xrprof) and for python, we can use [pyFlame](https://uwekorn.com/2018/10/05/pyflame.html) for this purpose.
```
johnkerl, line 86:
```suggestion
We decided to use [flamegraph](https://github.com/brendangregg/FlameGraph) as this common interface. The **_flamegraphs_** are a very popular interactive way of tracking performance metrics across software components. Therefore, the generic profiler will be given a set of custom profilers (and their arguments) to run and simply expect each profiler to generate a new **_flamegraph_** of the program software stack in a particular location and it adds the generated files to the database as well. Given the overhead of trace-based profilers, custom profilers are always going to be optional. For example upon detecting a regression in the generic profiler DB, we can rerun the application with the custom profiler to get the **_flamegraphs_** of the application. For R, we can use [xProf](https://github.com/atheriel/xrprof) and for python, we can use [pyFlame](https://uwekorn.com/2018/10/05/pyflame.html) for this purpose.
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+
+![alt_text](images/flamegraph.png "image_tooltip")
+
+### Benefits
+
+- A major benefit of this design is that the profilers (both generic and custom ones) are not necessarily targeted toward single cell applications and can be used for any services across CZI.
```
johnkerl, line 92:
```suggestion
- A major benefit of this design is that the profilers (both generic and custom ones) are not necessarily targeted toward SOMA and can be used for various services.
```

rfcs/profiler.md
```diff
@@ -0,0 +1,144 @@
...
+- Another benefit of this design is the fact that the generic profiler can be implemented and deployed in a reasonable amount of time opening the door for low-overhead perf/mem monitoring.
+- The third benefit of the design is that it can be easily extended by adding more custom profilers for different languages and software stack.
+
+### Drawbacks
+
+One drawback here is limiting custom profilers’ API to flamegraph. As mentioned earlier, there are many profilers with different output formats. One possible solution to this problem is to allow the byte array associated with each custom profiler in the DB schema to be open to different interpretations which while more scalable is a less secure solution.
```
johnkerl, line 98:
```suggestion
One drawback here is limiting custom profilers’ API to flamegraph. As mentioned earlier, there are many profilers with different output formats. One possible solution to this problem is to allow the byte array associated with each custom profiler in the DB schema to be open to different interpretations which, while more scalable, is a less secure solution.
```

johnkerl (changes requested):

ebezzi (approved):
LGTM. As Andrew suggested, I recommend to fill a more detailed implementation/deployment as a separate RFC or tech spec.

rfcs/profiler.md
```diff
@@ -0,0 +1,146 @@
+# Tech Spec: Performance Evaluation & Monitoring for Single Cell
```
johnkerl, line 1: The spec branding is SOMA. The impl branding is TileDB-SOMA. There is no "Single Cell" branding in our terminology. Please don't use this.

rfcs/profiler.md
johnkerl: @beroy please make the requested change

johnkerl (approved):

bkmartinjr:
Closing PR - the related work in TileDB-SOMA was not merged, and I believe was moved back to the CZI Census project repo.

PR #164, opened 2023-05-10
RFC on `soma_joinid`: identifier or index?

mlin:
Preview link: https://github.com/single-cell-data/SOMA/blob/mlin/rfc-on-r-soma-joinid/rfcs/soma_joinid_id_or_index.md


rfcs/soma_joinid_id_or_index.md
```diff
@@ -0,0 +1,25 @@
...
+
+Notice this defines `soma_joinid` merely as an *identifier*; not necessarily starting at zero or one, nor forming a contiguous range, nor even ascending with the data frame's row order. Those are properties associated with an *index*. Though the field is often populated like an index in practice, this need not be so. For example, we might create a `SOMAExperiment` by subsetting the variables of a larger existing one, preserving the original `soma_joinid` in the `var` data frame. Or, observables might be withdrawn from version to version of a `SOMAExperiment`, keeping the `soma_joinid` of the remaining entries in the `obs` data frame.
+
+**However**, beyond this primary definition, `soma_joinid` is also used as the row & column numbers of any `X` matrix in a `SOMAExperiment`. It clearly *is* an index in this secondary role, with the constraint that the row & column numbers correspond to the `soma_joinid` values in the `obs` and `var` dataframes.
```
johnkerl, line 9:
```suggestion
**However**, beyond this primary definition, `soma_joinid` is also used as the row and column numbers of any `X` matrix in a `SOMAExperiment`. It clearly *is* an index in this secondary role, with the constraint that the row & column numbers correspond to the `soma_joinid` values in the `obs` and `var` dataframes. 
```

johnkerl: If `X` is sparse then row and column numbers are not necessarily indices.

Only if it is dense do they necessarily become indices.

johnkerl (approved):

rfcs/soma_joinid_id_or_index.md
```diff
@@ -0,0 +1,25 @@
+# `soma_joinid`: identifier or index? (May 2023)
+
+The SOMA specification introduces `soma_joinid` like so:
+
+> Every `SOMADataFrame` must contain a column called `soma_joinid`, of type `int64` and domain `[0, 2^63-1]`. The `soma_joinid` column contains a unique value for each row in the `SOMADataFrame`, and intended to act as a join key for other objects, such as `SOMASparseNDArray`.
+
+Notice this defines `soma_joinid` merely as an *identifier*; not necessarily starting at zero or one, nor forming a contiguous range, nor even ascending with the data frame's row order. Those are properties associated with an *index*. Though the field is often populated like an index in practice, this need not be so. For example, we might create a `SOMAExperiment` by subsetting the variables of a larger existing one, preserving the original `soma_joinid` in the `var` data frame. Or, observables might be withdrawn from version to version of a `SOMAExperiment`, keeping the `soma_joinid` of the remaining entries in the `obs` data frame.
```
bkmartinjr, line 7: I find the definition of index confusing - there is nothing about an index that requires it to be contiguous, starting at zero/one, etc.

I think you want to separate the concept of an "index" (aka key) from "offset" (in Pandas terminology, you could have a string index, but you always have an integer `iloc` (offset)).

Using index in this way confuses the DB organization with the in-memory representation.

rfcs/soma_joinid_id_or_index.md
bkmartinjr, line 9: I disagree with this.  For sparse arrays (dataframe, sparsendarray), `soma_dim_N` is _not_ an index, in the way you define it above (i.e. a physical offset).  It is an identifier, just like the soma_joinid.   The only constraints on soma_joinid/soma_dim_N are that they are uint64.  They can be in any order, be non-contiguous, and span any range.

Common usage will use a contiguous range, but the spec absolutely (and intentionally) doesn't require that.

The one nuance here -- if you have a sparse X in an Experiment, you will only have _implicit_ zeros where there is a defined soma_joinid pair in the obs/var DataFrame.   In other words, any soma_dim_N not in the sparse array does not exist in the X index domain unless the joinid is defined in the corresponding obs/var dataframe.

Dense is another matter, and still has unresolved design issues due to the fact that we allow "sparse" joinid domains in the obs/var dataframes.

bkmartinjr:
ps. @mlin - re-reading my comments, this is complex enough topic that it may warrant a real-time conversation.  I'm available if you think this would be useful.

thetorpedodog:
two non–content-related notes from me (these apply to both open RFCs right now so I am pasting this note in both places):

1. Rebase so that you get the pre-commit format verification thingy
2. (Optional) Consider formatting to one sentence per line. I did this in 3b5891c678e9d08f4c8817cb6426f82c02675cf0, but I didn’t formally write it down, so it’s not a hard Rule but it is a format I have found useful for editing.

rfcs/soma_joinid_id_or_index.md
mlin: @bkmartinjr @johnkerl We need to find the precise terminology to convey the relationship between `soma_joinid` (not an index) with the row/column numbers of a matrix (obviously indexed sequences; whether the matrix is represented sparsely is an implementation detail, in my view). The original draft is too strong in *equating* them, admitted, but we need to convey that the treatment of `soma_joinid` as matrix row/column numbers in a certain context imposes certain constraints on its initial, broad definition -- that is probably the most important point of the whole essay.

jp-dark:
Closing as stale.

PR #165, opened 2023-05-15
Allow impl-specific `platform_config` values.

thetorpedodog:
This specifies that implementations may provide their own types to be passed as `platform_config`, and that other implementations should ignore things they don’t recognize.

---

Per the discussion at Friday’s meeting about allowing users to pass in rich objects, this is roughly what I was thinking would be the best approach. It’s pretty much what I said there, and tl;dr

1. Implementations can choose to accept their own rich types raw.
2. Other implementations should ignore ones they don't recognize and act as if nothing was passed.

I only vaguely remember who was most involved in that discussion, so please R+ anybody who seems relevant.

### DIFF
--- a/abstract_specification.md
+++ b/abstract_specification.md
@@ -1020,9 +1020,14 @@ All SOMA objects expose a `context` field that contain implementation-specific c
 
 ### Per-call configuration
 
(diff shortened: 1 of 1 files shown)


bkmartinjr (approved):
This all seems reasonable to me - thanks Paul!

johnkerl (approved):
🚢

thetorpedodog:
@pablo-gar bump, since this a spec change I would like all reviewers' approval. (If I misremembered the conversation and you didn’t have an opinion on this, you can remove yourself from the reviewers list (or approve if you think this is a good idea anyway).)

pablo-gar (approved):
Excellent attention to details and making the spec more specific

PR #166, opened 2023-05-19
Update `PlatformConfig` to allow impl-specific types

thetorpedodog:
This incorporates the specification changes from this commit: 89692959e9112d9cf1323cf0521a4cf0fa1efa49
into the Python implementation.

We restrict the multi-platform–compatible type to be specifically a ``dict`` (and not a ``Mapping``) so that if some implementation had a configuration that implemented Mapping, it would be excluded (i.e., we only look inside basic built-in types and no funny business).


johnkerl (approved):
LGTM

PR #169, opened 2023-07-26
[python] Add parameters to `ExperimentAxis` query methods

ebezzi:
Resolves https://github.com/single-cell-data/TileDB-SOMA/issues/1527
Testing will done on tiledbsoma (link TBD)

### DIFF
--- a/python-spec/src/somacore/query/query.py
+++ b/python-spec/src/somacore/query/query.py
@@ -192,6 +192,7 @@ class ExperimentAxisQuery(Generic[_Exp]):
     def X(
         self,
         layer_name: str,
+        *,
         batch_size: options.BatchSize = options.BatchSize(),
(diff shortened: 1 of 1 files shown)


johnkerl (approved):

python-spec/src/somacore/query/query.py
```diff
@@ -166,11 +189,24 @@ def indexer(self) -> "AxisIndexer":
         """
         return self._indexer
 
-    def X(self, layer_name: str) -> data.SparseRead:
+    def X(
+        self,
+        layer_name: str,
+        batch_size: options.BatchSize = options.BatchSize(),
```
thetorpedodog, line 195: Maybe throw a `*` here, so the rest of these are kw-only?

PR #173, opened 2023-09-29
Update CI-related stuff to new versions.

thetorpedodog:
This updates our generated requirements.txt files and the various tools we use in pre-commit to the latest versions. There is no runtime change; this only affects what we use in testing and static analysis.


thetorpedodog:
(this is also intended to shut up a dependabot thing)

bkmartinjr (approved):

Issue #177, opened 2023-10-09
`ExperimentAxisQuery._read()` and `ExperimentAxisQuery.to_anndata()` should read and export, respectively, `obsm` `varm` `obsp` and `varp`.


pablo-gar:
Currently if a SOMA `Experiment` has arrays in `obsm` `varm` `obsp` and `varp` they are not read by `ExperimentAxisQuery._read()` and thus not exported by `ExperimentAxisQuery.to_anndata()`.

These arrays should be able to be exported to an AnnData via  `ExperimentAxisQuery.to_anndata()`. 

This ticket doesn't cover the inclusion of the methods `.obsm()` `obsp()` `varm` and `varp()`

johnkerl:
@ebezzi does this include both #178 and #179?

PR #178, opened 2023-10-13
Provide obsm/varm from ExperimentAxisQuery, as obsp/varp

johnkerl:
WIP

Context: https://github.com/single-cell-data/SOMA/issues/177


johnkerl:
Replaced by #179

PR #179, opened 2023-11-06
Add obsm, etc to ExperimentAxisQuery

ebezzi:
Adds `obsm` (for now) to `ExperimentAxisQuery._read` and `.to_anndata`. Tested manually so far.

This is a draft to validate the approach. If it looks OK, I will:
1. Finish implementing obsp, varm, varp
2. Add tests in TileDB-SOMA
3. Make this a bit more DRY.

### DIFF
--- a/python-spec/src/somacore/query/query.py
+++ b/python-spec/src/somacore/query/query.py
@@ -240,6 +240,18 @@ class ExperimentAxisQuery(Generic[_Exp]):
         Lifecycle: maturing
         """
         return self._axisp_inner(_Axis.VAR, layer)
+    
+    def obsm(self, layer: str) -> data.SparseRead:
+        """Returns an ``obsm`` layer as a sparse read.
+        Lifecycle: maturing
+        """
+        return self._axism_inner(_Axis.OBS, layer)
+
+    def varm(self, layer: str) -> data.SparseRead:
+        """Returns an ``varm`` layer as a sparse read.
+        Lifecycle: maturing
+        """
+        return self._axism_inner(_Axis.VAR, layer)
 
     def to_anndata(
         self,
@@ -247,6 +259,7 @@ class ExperimentAxisQuery(Generic[_Exp]):
         *,
         column_names: Optional[AxisColumnNames] = None,
         X_layers: Sequence[str] = (),
+        obsm_keys: Sequence[str] = [],
     ) -> anndata.AnnData:
         """
         Executes the query and return result as an ``AnnData`` in-memory object.
@@ -264,6 +277,7 @@ class ExperimentAxisQuery(Generic[_Exp]):
             X_name,
             column_names=column_names or AxisColumnNames(obs=None, var=None),
             X_layers=X_layers,
+            obsm_keys=obsm_keys,
         ).to_anndata()
 
     # Context management
@@ -305,6 +319,7 @@ class ExperimentAxisQuery(Generic[_Exp]):
         *,
         column_names: AxisColumnNames,
         X_layers: Sequence[str],
+        obsm_keys: Sequence[str] = [], # TODO: Add obsp_keys, varm_keys, varp_keys
     ) -> "_AxisQueryResult":
         """Reads the entire query result into in-memory Arrow tables.
 
@@ -341,8 +356,18 @@ class ExperimentAxisQuery(Generic[_Exp]):
             for _xname in all_x_arrays
         }
 
+        # TODO: do it in parallel?
+        obsm = dict()
+        for key in obsm_keys:
+            obsm[key] = self._axism_inner_csr(_Axis.OBS, key)
+            # _obsm = self.obsm(key)
+            # obsm[key] = _obsm
+            # _, n_cols = _obsm.shape
+            # col_idx = pa.array(range(n_cols), type=pa.uint32())
+            # obsm[key] = _fast_csr.read_scipy_csr(_obsm, self.obs_joinids(), col_idx)
+
         x = x_matrices.pop(X_name)
-        return _AxisQueryResult(obs=obs_table, var=var_table, X=x, X_layers=x_matrices)
+        return _AxisQueryResult(obs=obs_table, var=var_table, X=x, obsm=obsm, X_layers=x_matrices)
 
     def _read_both_axes(
         self,
@@ -433,6 +458,47 @@ class ExperimentAxisQuery(Generic[_Exp]):
 
         joinids = getattr(self._joinids, axis.value)
         return axisp[layer].read((joinids, joinids))
+    
+    def _axism_inner(
+        self,
+        axis: "_Axis",
+        layer: str,
+    ) -> data.SparseRead:
+        key = axis.value + "m"
+
+        if key not in self._ms:
+            raise ValueError(f"Measurement does not contain {key} data")
+
+        axism = self._ms.obsm if axis is _Axis.OBS else self._ms.varm
+        if not (layer and layer in axism):
+            raise ValueError(f"Must specify '{key}' layer")
+        
+        n_row, n_col = axism[layer].shape
+        col_joinids = pa.array(range(n_col))
+
+        joinids = getattr(self._joinids, axis.value)
+        return axism[layer].read((joinids, col_joinids))
+    
+    def _axism_inner_csr(
+        self,
+        axis: "_Axis",
+        layer: str,
+    ) -> data.SparseRead:
+        key = axis.value + "m"
+
+        if key not in self._ms:
+            raise ValueError(f"Measurement does not contain {key} data")
+
+        axism = self._ms.obsm if axis is _Axis.OBS else self._ms.varm
+        if not (layer and layer in axism):
+            raise ValueError(f"Must specify '{key}' layer")
+        
+        _, n_col = axism[layer].shape
+        col_idx = pa.array(range(n_col), type=pa.int64())
+
+        joinids = getattr(self._joinids, axis.value)
+
+        return _fast_csr.read_scipy_csr(axism[layer], joinids, col_idx)
 
     @property
     def _obs_df(self) -> data.DataFrame:
@@ -472,6 +538,14 @@ class _AxisQueryResult:
     """Experiment.ms[...].X[...] query slice, as an SciPy sparse.csr_matrix """
     X_layers: Dict[str, sparse.csr_matrix] = attrs.field(factory=dict)
     """Any additional X layers requested, as SciPy sparse.csr_matrix(s)"""
+    obsm: Dict[str, sparse.csr_matrix] = attrs.field(factory=dict)
+    """Experiment.obsm query slice, as SciPy sparse.csr_matrix(s)"""
+    obsp: Dict[str, sparse.csr_matrix] = attrs.field(factory=dict)
+    """Experiment.obsp query slice, as SciPy sparse.csr_matrix(s)"""
+    varm: Dict[str, sparse.csr_matrix] = attrs.field(factory=dict)
+    """Experiment.varm query slice, as SciPy sparse.csr_matrix(s)"""
+    varp: Dict[str, sparse.csr_matrix] = attrs.field(factory=dict)
+    """Experiment.varp query slice, as SciPy sparse.csr_matrix(s)"""
 
     def to_anndata(self) -> anndata.AnnData:
         obs = self.obs.to_pandas()
@@ -481,7 +555,14 @@ class _AxisQueryResult:
         var.index = var.index.astype(str)
 
         return anndata.AnnData(
-            X=self.X, obs=obs, var=var, layers=(self.X_layers or None)
+            X=self.X, 
+            obs=obs, 
+            var=var, 
+            obsm=(self.obsm or None), 
+            obsp=(self.obsp or None), 
+            varm=(self.varm or None), 
+            varp=(self.varp or None), 
+            layers=(self.X_layers or None)
         )
 
 


python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +456,47 @@ def _axisp_inner(
...
+        if not (layer and layer in axism):
+            raise ValueError(f"Must specify '{key}' layer")
+
+        n_row, n_col = axism[layer].shape
+        col_joinids = pa.array(range(n_col))
+
+        joinids = getattr(self._joinids, axis.value)
+        return axism[layer].read((joinids, col_joinids))
+
+    def _axism_inner_csr(
```
ebezzi, line 479: This method exists purely to be passed to `_read`, which loads everything in memory. This works for now but I am not sure this is the right format we want for `obsm`. In principle, for `obsm` it would be better to use a csc matrix since it has way more rows than column. Most embeddings though are dense, so maybe we could just use a numpy array here?

python-spec/src/somacore/query/query.py
```diff
@@ -241,12 +241,25 @@ def varp(self, layer: str) -> data.SparseRead:
...
+        """
+        return self._axism_inner(_Axis.VAR, layer)
+
     def to_anndata(
         self,
         X_name: str,
         *,
         column_names: Optional[AxisColumnNames] = None,
         X_layers: Sequence[str] = (),
+        obsm_keys: Sequence[str] = [],
```
bkmartinjr, line 262: Need to use consistent names for X and obsm:
* `X_layers`
* `obsm_keys`

Suggest simply using `_layers`

Also, use a tuple, not a list, as the default, a la `X_layers`

bkmartinjr: Also, missing the rest of the layers (varm, etc).  Any reason we should not add all of it in one pass?

ebezzi: Before I add the other layers, I'd like the approach to be validated

python-spec/src/somacore/query/query.py
```diff
@@ -305,6 +319,7 @@ def _read(
         *,
         column_names: AxisColumnNames,
         X_layers: Sequence[str],
+        obsm_keys: Sequence[str] = [],  # TODO: Add obsp_keys, varm_keys, varp_keys
     ) -> "_AxisQueryResult":
         """Reads the entire query result into in-memory Arrow tables.
```
bkmartinjr, line 324: This routine _used_ to only return Arrow objects.  At some point, it looks like it was modified to return SciPy objects.

(looks like this was from a PR landed by @thetorpedodog )

We should at least clean up the comments so they match the code. I'm not sure if there are other implications, other than this routine being less general (i.e., it is now only useful in the to_anndata code path, whereas it was _in theory_ more general previously)

python-spec/src/somacore/query/query.py
```diff
@@ -341,8 +356,15 @@ def _read(
             for _xname in all_x_arrays
         }
 
+        # TODO: do it in parallel?
```
bkmartinjr, line 359: Perhaps yes, as it should help with load perf, but I'd get the code structure and APIs nailed down first....

python-spec/src/somacore/query/query.py
bkmartinjr: I suggest looking at the AnnData documentation.  obsm/varm/obsp/varp are ndarray in typical use, and I don't see any reason they would not be here...  Using a sparse matrix is likely to create incompatibility with downstream user assumptions.

python-spec/src/somacore/query/query.py
```diff
@@ -341,8 +356,15 @@ def _read(
...
+        obsm = dict()
+        for key in obsm_keys:
+            obsm[key] = self._axism_inner_csr(_Axis.OBS, key)
```
bkmartinjr, line 362: why is this a CSR - it should be an ndarray... 

See [AnnData obsm doc](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.obsm.html#anndata.AnnData.obsm)

python-spec/src/somacore/query/query.py
bkmartinjr: As a second thought:  if this code path is dedicated to `to_anndata`, we should name and document it as such.

I haven't given it a huge amount of thought, but I am not opposed to this in principle.  In theory it is a private API, which we can mod at will

python-spec/src/somacore/query/query.py
```diff
@@ -472,6 +535,14 @@ class _AxisQueryResult:
     """Experiment.ms[...].X[...] query slice, as an SciPy sparse.csr_matrix """
     X_layers: Dict[str, sparse.csr_matrix] = attrs.field(factory=dict)
     """Any additional X layers requested, as SciPy sparse.csr_matrix(s)"""
+    obsm: Dict[str, sparse.csr_matrix] = attrs.field(factory=dict)
```
bkmartinjr, line 538: Originally, `_AxisQueryResult` structure was pure Arrow. If it is now a mismash of stuff for use in (only) `to_anndata`, we should rename and clean it up so that it is so-documented.

bkmartinjr (changes requested):
several concerns:
* this doesn't produce a "legal" AnnData, which specifies that obsm/obsp/varm/varp are ndarray
* the purpose/signature of query._read seems to morphed over time (not just from this PR), such that the docstrings and sig don't match the original "produces pure Arrow objects" intent.  We should make an intentional decision to either a) dedicate this code path to the to_anndata use case exclusively, or b) refactor it back to generating Arrow objects, with the conversion to AnnData wrapped around it.  I'm OK with either.

python-spec/src/somacore/query/query.py
ebezzi: Good point, for some reason I thought AnnData didn't enforce the typing here, but it looks like it does. I'll convert to ndarray.

bkmartinjr: its more than "enforce typing" - it is conceptually a dense array in the AnnData data model (but not in SOMA)

python-spec/src/somacore/query/query.py
ebezzi: By the way, it looks like AnnData uses ["keys"](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.obsm.html) to define the members of obsm, etc. [layers](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.layers.html) seems to specifically refer to X. I'll leave the naming as is right now, but open to more discussion.

python-spec/src/somacore/query/query.py
ebezzi: I changed the docs and remove the part where it mentions Arrow objects. We can take one step further and make the class 100% specific to `to_anndata`.

python-spec/src/somacore/query/query.py
```diff
@@ -241,12 +241,28 @@ def varp(self, layer: str) -> data.SparseRead:
         """
         return self._axisp_inner(_Axis.VAR, layer)
 
+    def obsm(self, layer: str) -> data.SparseRead:
+        """Returns an ``obsm`` layer as a sparse read.
+        Lifecycle: maturing
```
bkmartinjr, line 246: I think these should be `experimental` until they have been in a release or two.  They can't be maturing if they haven't even been merged :-)

python-spec/src/somacore/query/query.py
```diff
@@ -241,12 +241,28 @@ def varp(self, layer: str) -> data.SparseRead:
         """
         return self._axisp_inner(_Axis.VAR, layer)
 
+    def obsm(self, layer: str) -> data.SparseRead:
```
bkmartinjr, line 244: Argument names are inconsistent between `obsm`, etc and `to_anndata` - one uses `layer`, the other uses `keys`

Noting in case this is not intentional.  IMHO, they would ideally use the same names.   Collection - uses `key`

python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +501,81 @@ def _axisp_inner(
...
+    ) -> data.SparseRead:
+        key = axis.value + "m"
+
+        if key not in self._ms:
+            raise ValueError(f"Measurement does not contain {key} data")
+
+        axism = self._ms.obsm if axis is _Axis.OBS else self._ms.varm
+        if not (layer and layer in axism):
+            raise ValueError(f"Must specify '{key}' layer")
+
```
bkmartinjr, line 517: _axisp_inner contains an additional check for the type of the object (SparseNDArray). is there a reason to not do a check here, given that the subsequent code requires that it be a SparseNDArray?

python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +501,81 @@ def _axisp_inner(
         joinids = getattr(self._joinids, axis.value)
         return axisp[layer].read((joinids, joinids))
 
+    def _axism_inner(
```
bkmartinjr, line 504: this code is almost identical to _axisp_inner.  Can we DRY?

python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +501,81 @@ def _axisp_inner(
...
+
+        if key not in self._ms:
+            raise ValueError(f"Measurement does not contain {key} data")
+
+        axism = self._ms.obsm if axis is _Axis.OBS else self._ms.varm
+        if not (layer and layer in axism):
+            raise ValueError(f"Must specify '{key}' layer")
+
+        n_row, n_col = axism[layer].shape
+        col_joinids = pa.array(range(n_col))
```
bkmartinjr, line 519: why not a slice or None (which is synonym for slice(None))?

python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +501,81 @@ def _axisp_inner(
...
+
+        joinids = getattr(self._joinids, axis.value)
+        return axism[layer].read((joinids, col_joinids))
+
+    def _axisp_inner_ndarray(
```
bkmartinjr, line 524: all of these _inner_ndarray methods are largely identical to the _inner methods, with some processing done on the SparseRead result.

They should be DRY'ed.  For example, this one (I think) could be something like:

```
T = self._axisp_inner(axis, layer)
n_row = n_col = len(self._joinids.obs) if axis is _Axis.OBS else len(self._joinids.var)
idx = (self.indexer.by_obs if axis is _Axis.OBS else self.indexer.by_var)(T["soma_dim_0"])
Z = np.zeros(n_row * n_col, dtype=np.float32)
np.put(Z, idx * n_col + T["soma_dim_1"], T["soma_data"])
return Z.reshape(n_row, n_col)
```

python-spec/src/somacore/query/query.py
```diff
@@ -305,8 +333,12 @@ def _read(
         *,
         column_names: AxisColumnNames,
         X_layers: Sequence[str],
+        obsm_keys: Sequence[str] = (),
+        obsp_keys: Sequence[str] = (),
+        varm_keys: Sequence[str] = (),
+        varp_keys: Sequence[str] = (),
     ) -> "_AxisQueryResult":
-        """Reads the entire query result into in-memory Arrow tables.
+        """Reads the entire query result in memory.
```
bkmartinjr, line 341: we still need to sort this out, as it is a hodge podge of formats.   If it is only for use in the anndata code path, it really should return Pandas Dataframes for obs/var, etc.

If it is for other uses, it should return chunked arrays, not ndarray for obsm, etc.  (ie.., avoid doing the extra copy that ndarray implies).

Either way, lets pick a purpose for this routine and have it fulfill that destiny.  At the moment, it is a mish-mash.

This includes both this routine and its return type.

I'm OK with either approach, but note that Arrow-native types are more useful, and give the caller the choice of doing zero-copy or not (the current code forces a copy in some cases)

python-spec/src/somacore/query/query.py
bkmartinjr: This needs resolution. There is some risk of over-parallizing, and it will use a bit more memory to do concurrently.  But it will also be quite a bit faster given the reindexing implicit in each of these reads.

My gut feeling (not benchmarked) is to do them all concurrently (e.g., submit to the shared query thread pool).

Curious what @thetorpedodog and @johnkerl think.

bkmartinjr (changes requested):
Primary ask: there is a lot of cut&pasted code.  DRY. 
Also need to resolve the issue of concurrent reading of obsm/....

python-spec/src/somacore/query/query.py
ebezzi: I would settle for `keys`, since that's how AnnData calls them, plus `layers` don't seem to make a lot of sense for embeddings or pairwise matrices (for one, embeddings can have a different number of columns therefore they don't even overlap). The issue is that `obsp` and `varp` were already part of the spec, so renaming the variable to `keys` would be a breaking change. Worth the "risk"?

python-spec/src/somacore/query/query.py
ebezzi: Good call out. I originally removed it because I was planning to support `DenseNDArray`s as well, but we ended up not doing it. Adding it back.

python-spec/src/somacore/query/query.py
ebezzi: At this stage, given that `_read` is a private API and is only used by `to_anndata`, I am proposing that we just merge the two methods. If in the future we'll introduce support for some other data format that can feed directly from Arrow objects, we can always go back and add back a properly defined `_read` method that is more general, and move the conversions to pandas/numpy/scipy back to the `to_anndata`. Alternatively we can just keep the two methods separate and push the conversion to pandas Dataframes up to `_read`.

python-spec/src/somacore/query/query.py
ebezzi: Should be way more DRY now.

python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +501,59 @@ def _axisp_inner(
...
+        if not (layer and layer in axism):
+            raise ValueError(f"Must specify '{key}' layer")
+        
+        if not isinstance(axism[layer], data.SparseNDArray):
+            raise TypeError(f"Unexpected SOMA type stored in '{key}' layer")
+
+        joinids = getattr(self._joinids, axis.value)
+        return axism[layer].read((joinids, slice(None)))
+    
+    def _convert_to_ndarray(self, is_obs: bool, T: pa.Table, n_row: int, n_col: int) -> np.ndarray:
```
bkmartinjr, line 524: I'm fairly sure that this has a bug -- it always reindexes the obs joinids, even if axis is VAR.  And it calculates coordinates with the var dimension even when it shouldn't 

you probably want something like:
```
major_dim = 0 if is_obs else 1
minor_dim = 1-major_dim
idx = (self.indexer.by_obs if is_obs else self.indexer.by_var)(T[f"soma_dim_{major_dim}"])
Z = np.zeros(n_row * n_col, dtype=np.float32)
np.put(Z, idx * n_col + T[f"soma_dim_{minor_dim}"], T["soma_data"])
return Z.reshape(n_row, n_col)
```

bkmartinjr: oh, and good to ensure that the unit tests are sufficient to capture this bug ;-)

python-spec/src/somacore/query/query.py
bkmartinjr: I don't have an opinion on the proposed merge (other than to note that giant methods are hard to follow).  I'm just saying that if it is for to_anndata, you might as well return AnnData-compat types (e.g., Pandas), etc.

this is a minor point - just seems like we have a half-modified API signature, which is worth cleaning up.

python-spec/src/somacore/query/query.py
bkmartinjr: If you want to propose a cleanup, e can fairly easily support both, deprecate one, and remove it in a few releases.

python-spec/src/somacore/query/query.py
bkmartinjr: update:  reviewer (me) confusion. Apologies.

python-spec/src/somacore/query/query.py
thetorpedodog: If I’m following this discussion correctly, is the concern about renaming the parameters to obsp/varp to `key`? This would *technically* be a break, but only affect anybody who was writing `whatever.varp(layer=xxx)`. I feel pretty comfortable saying this would be fine even in the absence of a transition period.

If we were really concerned, we could also make it a positional-only argument (i.e., use something like `__layer` for the name in old Python, or `def varm(self, /, key: str)` once we jettison v3.7, and end naming worries once and for all.

If I’m misreading, feel free to ignore whatever I just said.

python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +501,59 @@ def _axisp_inner(
...
+        np.put(Z, idx * n_col + T["soma_dim_1"], T["soma_data"])
+        return Z.reshape(n_row, n_col)
+
+    def _axisp_inner_ndarray(
+        self,
+        axis: "_Axis",
+        layer: str,
+    ) -> np.ndarray:
+        is_obs = axis is _Axis.OBS
+        n_row = n_col = len(self._joinids.obs) if is_obs else len(self._joinids.var)
```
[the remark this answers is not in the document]
thetorpedodog: decided to see what this would look like here https://github.com/single-cell-data/SOMA/pull/183

python-spec/src/somacore/query/query.py
ebezzi: Yes, the discussion was about renaming `layer` to `key`. Given that `layer` was, IMHO, a pretty confusing name in the first place for obsp/varp, I'm proposing we just introduce the breaking change and rename it to `key`. @johnkerl any concerns from you?

python-spec/src/somacore/query/query.py
ebezzi: I ended up making sure that the `_AxisQueryResult` object only has "final" types (i.e. dataframes, ndarray, scipy.sparse) and fixed the docstrings a bit. Let me know if we want to make the fact that this is - effectively - a proxy to an AnnData object more clear in the docs.

python-spec/src/somacore/query/query.py
```diff
@@ -341,8 +381,39 @@ def _read(
             for _xname in all_x_arrays
         }
 
+        def _read_axis_mappings(fn, axis, keys: Sequence[str]) -> Dict[str, np.ndarray]:
+            return {key: fn(axis, key) for key in keys}
+
+        obsm_ft = self._threadpool.submit(
```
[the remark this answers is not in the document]
bkmartinjr: I'm fine with this solution, but I would submit these tasks before you start the X read, so they all run in parallel.  Otherwise, the read of X and the read of the obsm/... are serialized

python-spec/src/somacore/query/query.py
```diff
@@ -434,6 +512,62 @@ def _axisp_inner(
...
+    ) -> data.SparseRead:
+        key = axis.value + "m"
+
+        if key not in self._ms:
+            raise ValueError(f"Measurement does not contain {key} data")
+
+        axism = self._ms.obsm if axis is _Axis.OBS else self._ms.varm
+        if not (layer and layer in axism):
+            raise ValueError(f"Must specify '{key}' layer")
+
```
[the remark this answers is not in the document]
ebezzi: I did run pre-commit and it passes for me... is it possible that you didn't pull the latest commits?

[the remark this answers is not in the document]
bkmartinjr: confirmed - all is now good.

python-spec/src/somacore/query/query.py
```diff
@@ -342,8 +399,29 @@ def _read(
             for _xname in all_x_arrays
         }
 
+        obsm = obsm_ft.result()
+        obsp = obsp_ft.result()
+        varm = varm_ft.result()
+        varp = varp_ft.result()
```
thetorpedodog, line 405: It seems like these could be passed as the named arguments to `_AxisQueryResult` without the need for a temporary variable.

thetorpedodog (approved):
A few minor style suggestions; nothing critical. Looks good!

johnkerl (commented):
🚢

python-spec/src/somacore/query/query.py
```diff
@@ -333,6 +374,22 @@ def _read(
...
             all_x_arrays[_xname] = x_array
 
+        def _read_axis_mappings(fn, axis, keys: Sequence[str]) -> Dict[str, np.ndarray]:
+            return {key: fn(axis, key) for key in keys}
+
+        obsm_ft = self._threadpool.submit(
+            _read_axis_mappings, self._axism_inner_ndarray, _Axis.OBS, obsm_keys
+        )
+        obsp_ft = self._threadpool.submit(
+            _read_axis_mappings, self._axisp_inner_ndarray, _Axis.OBS, obsp_keys
```
pablo-gar, line 384: @bkmartinjr and @ebezzi I believe you had a conversation about this offline (or maybe here but I don't see it). We know that the most common uses case of axis**p** arrays are numerically sparse, and even though anndata's schema says they should be numpy dense arrays, scanpy's methods fill it in with scipy sparse matrices.

Ultimately we would like to move to a world where either AnnData schema's takes both (dense and sparse) or only sparse. Given that the ecosystem already violates the schema in favor of better numerical representation I lean towards tiledbsoma also violating the schema, and then we request AnnData's schema to be relaxed.

What are your thoughts?

bkmartinjr: In principle I have no issues.  You might want to ping Isaac V. and see what he thinks

pablo-gar: Thanks I wanted to make sure I didn't miss anything important. I don't see it as a blocker for now, I will file a few issues (here and in AnnData) to move towards the support of sparse matrices in the axisp arrays.

pablo-gar (approved):
LGTM from an API signature viewpoint!

aaronwolen (approved):
🙏🏻

PR #181, opened 2023-11-14
Allow overriding of CSR accumulators with an `IndexLike`

thetorpedodog:
To support a custom indexer in TileDB-SOMA, this introduces a basic abstraction that is used to build the Indexer used by the CSR Accumulator.  This can then be easily overridden by an implementation's custom subclass to swap that out without having to duplicate substantial parts of the code.

The naming is, admittedly, not ideal; this is in part due to the naming of the things we're trying to abstract over (the Pandas `Index` type which has the `get_indexer` method).

---

This is the somacore counterpart of https://github.com/single-cell-data/TileDB-SOMA/pull/1728. It *is* entirely compatible with current code, though; once we’re positive that it fufills the needs of the tiledb-soma reindexer, we can merge it and release it and tiledbsoma installations will continue to work fine.

### DIFF
--- a/python-spec/src/somacore/query/types.py
+++ b/python-spec/src/somacore/query/types.py
@@ -1,0 +1,33 @@
+"""Common types used across SOMA query modules."""
+
+from typing import Any, Callable
+
+import numpy as np
+import numpy.typing as npt
+from typing_extensions import Protocol
+
+
+class IndexLike(Protocol):
+    """The basics of what we expect an Index to be.
+
+    This is a basic description of the parts of the ``pandas.Index`` type
+    that we use. It is intended as a rough guide so an implementor can know
+    that they are probably passing the right "index" type into a function,
+    not as a full specification of the types and behavior of ``get_indexer``.
+    """
+
+    def get_indexer(
+        self,
+        target: npt.NDArray[np.int64],
+        method: object = ...,
+        limit: object = ...,
+        tolerance: object = ...,
+    ) -> Any:
+        """Something compatible with Pandas' Index.get_indexer method."""
+
+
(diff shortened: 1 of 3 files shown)


bkmartinjr (approved):
I did a code read, and this looks great to me.  Thanks for cleaning up some of the crufty code!

beroy:
Plz fix the formattings

python-spec/src/somacore/query/types.py
```diff
@@ -26,7 +26,7 @@ def get_indexer(
         """Something compatible with Pandas' Index.get_indexer method."""
 
 
-IndexFactory = Callable[[npt.NDArray[np.int64]], "IndexLike"]
+IndexFactory = Callable[[npt.NDArray[np.int64], Optional[Any]], "IndexLike"]
```
thetorpedodog, line 29: The reason we have this take the NDArray as the only parameter is so that we can use the function `pandas.Index` as an index factory. With this added, `pandas.Index` no longer works, which is why the format check is now failing.

beroy:
@thetorpedodog can you plz merge this as my C++ reindexer is dependent on this?

beroy (approved):

thetorpedodog:
I would like to wait until we get approval on the TileDB-SOMA side of the house. I am _pretty_ confident that this is going to be it, but I don’t want to cut a release prematurely. It will be very quick—after merging this, we can immediately make the tag and everything will be ready in a matter of minutes. (There will be no rush to release TileDB-SOMA; this is fully compatible with existing TileDB-SOMA.)

thetorpedodog:
We’re good to go!

PR #182, opened 2023-11-17
Update pyarrow dependency

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/issues/1926

### DIFF
--- a/python-spec/src/somacore/collection.py
+++ b/python-spec/src/somacore/collection.py
@@ -1,11 +1,7 @@
 import abc
 from typing import Any, MutableMapping, Optional, Sequence, Type, TypeVar, overload
 
-# ruff: noqa
 import pyarrow as pa
-
-# ruff: noqa
-import pyarrow_hotfix
 from typing_extensions import Final, Self
 
 from . import base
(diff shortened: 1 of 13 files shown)


thetorpedodog:
In this case, the `requirements.txt` files are intended to specify the *exact* version of every library that should be installed in the test environment; they can be automatically updated with `./python-spec/update-requirements-txt`.

johnkerl:
Commit https://github.com/single-cell-data/SOMA/pull/182/commits/331d5a79836b942f3928832e4bcf4084898ff450 implements the plan at https://github.com/single-cell-data/TileDB-SOMA/issues/1926#issuecomment-1821208184

python-spec/src/somacore/collection.py
```diff
@@ -1,7 +1,11 @@
 import abc
 from typing import Any, MutableMapping, Optional, Sequence, Type, TypeVar, overload
 
+# ruff: noqa
 import pyarrow as pa
+
+# ruff: noqa
+import pyarrow_hotfix
```
thetorpedodog, line 8: This only needs to be included once in `somacore/__init__.py`, since importing any somacore module necessarily entails importing that file. Putting the lint silence on the same line:

```python
import pyarrow_hotfix  # noqa: F401
```

will ensure minimal lint-disabling and messing around with the formatting.

johnkerl: https://github.com/single-cell-data/SOMA/pull/182/commits/1c2b483ba4a0067bf83b5312612d96812b922d24

thetorpedodog (approved):

PR #183, opened 2023-11-30
Add `_Axis.getattr_from` and `_Axis.getitem_from`.

thetorpedodog:
This aims to eliminate the annoyance (and potential error) of writing

    thing = whatever.obs if axis is _Axis.OBS else whatever.var

by letting you say

    thing = axis.getattr_from(whatever)

instead.

### DIFF
--- a/python-spec/src/somacore/query/query.py
+++ b/python-spec/src/somacore/query/query.py
@@ -10,6 +10,7 @@ from typing import (
     Tuple,
     TypeVar,
     Union,
+    overload,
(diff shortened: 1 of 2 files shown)


bkmartinjr (approved):
OK w/ me as long as we are vigilant about using it!

Afterthought: I wonder if the name could be more self-explanatory?  E.g.,  `get_axis_attr` or some such?

ebezzi (approved):

PR #184, opened 2024-01-10
ExperimentAxisQuery uses the thread pool from ContextBase 

ebezzi:
Companion PR to: https://github.com/single-cell-data/TileDB-SOMA/pull/2001

### DIFF
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -10,7 +10,7 @@ from typing import Any, ClassVar, MutableMapping, Optional
 from typing_extensions import LiteralString, Self
 
 from . import options
-from .types import ContextBase
+from . import types
 
 
 class SOMAObject(metaclass=abc.ABCMeta):
@@ -25,7 +25,8 @@ class SOMAObject(metaclass=abc.ABCMeta):
         uri: str,
         mode: options.OpenMode = "r",
         *,
-        context: Optional[Any] = None,
+        context: Optional[Any] = None
+        ,
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> Self:
         """Opens the SOMA object of this type at the given URI.
@@ -66,7 +67,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
         raise NotImplementedError()
 
     @property
-    def context(self) -> Optional[ContextBase]:
+    def context(self) -> Optional[types.ContextBase]:
         """A value storing implementation-specific configuration information.
 
         This contains long-lived (i.e., not call-specific) information that is
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -25,8 +25,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
         uri: str,
         mode: options.OpenMode = "r",
         *,
-        context: Optional[Any] = None
-        ,
+        context: Optional[Any] = None,
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> Self:
         """Opens the SOMA object of this type at the given URI.
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -10,6 +10,7 @@ from typing import Any, ClassVar, MutableMapping, Optional
 from typing_extensions import LiteralString, Self
 
 from . import options
+from .types import ContextBase
 
 
 class SOMAObject(metaclass=abc.ABCMeta):
@@ -24,7 +25,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
         uri: str,
         mode: options.OpenMode = "r",
         *,
-        context: Optional[Any] = None,
+        context: Optional[ContextBase] = None,
         platform_config: Optional[options.PlatformConfig] = None,
     ) -> Self:
         """Opens the SOMA object of this type at the given URI.
@@ -42,7 +43,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
 
     @classmethod
     @abc.abstractmethod
-    def exists(cls, uri: str, *, context: Optional[Any] = None) -> bool:
+    def exists(cls, uri: str, *, context: Optional[ContextBase] = None) -> bool:
         """Checks whether a SOMA object of this type is stored at the URI.
 
         Args:
@@ -65,7 +66,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
         raise NotImplementedError()
 
     @property
-    def context(self) -> Any:
+    def context(self) -> Optional[ContextBase]:
         """A value storing implementation-specific configuration information.
 
         This contains long-lived (i.e., not call-specific) information that is
--- a/python-spec/src/somacore/base.py
+++ b/python-spec/src/somacore/base.py
@@ -25,7 +25,7 @@ class SOMAObject(metaclass=abc.ABCMeta):
         uri: str,
         mode: options.OpenMode = "r",
         *,
-        context: Optional[ContextBase] = None,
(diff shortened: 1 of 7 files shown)


python-spec/src/somacore/query/query.py
```diff
@@ -587,11 +588,13 @@ def _var_df(self) -> data.DataFrame:
 
     @property
     def _threadpool(self) -> futures.ThreadPoolExecutor:
-        """Creates a thread pool just in time."""
+        """
+        Returns the threadpool provided by the experiment's context.
+        If not available, creates a thread pool just in time."""
+        if self.experiment.context._threadpool is not None:
```
bkmartinjr, line 594: the Experiment-ish protocol types _threadpool as a required value, i.e., won't be None.  Which is correct?

bkmartinjr (changes requested):
One typing nit - otherwise looks good.

python-spec/src/somacore/types.py
```diff
@@ -75,3 +76,12 @@ def is_slice_of(__obj: object, __typ: Type[_T]) -> TypeGuard[Slice[_T]]:
...
     )
+
+
+class ContextBase(Protocol):
+    """A protocol for a context manager that can be used as a base class.
+
+    The only requirement for somacore is that it should contain a threadpool.
+    """
+
+    _threadpool: futures.ThreadPoolExecutor
```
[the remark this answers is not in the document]
ebezzi: Good call, I switched the context to be `Optional` (as it is required by the definition of `Experiment`). For the threadpool, it can be either. The current implementation forces the threadpool to be not None, but I think it makes sense to leave that Optional as well, in case any other implementation prefers not to provide a thread pool. Regardless, the ExperimentAxisQuery code has a fallback path that manages its own threadpool in case either the context or the pool is missing.

python-spec/src/somacore/query/query.py
ebezzi: See comment above - I put all of them as `Optional`.

thetorpedodog:
If we're going to do this, we should also change the places where it is accepted in parameters / returned from accessors to be this base context protocol.

I also feel like the `threadpool` member should be public (i.e., un-prefixed), since it is intended for semi-public consumption (within SOMA libraries, but still).

ebezzi:
> If we're going to do this, we should also change the places where it is accepted in parameters / returned from accessors to be this base context protocol.
> 
> I also feel like the `threadpool` member should be public (i.e., un-prefixed), since it is intended for semi-public consumption (within SOMA libraries, but still).

Good points. I implemented both. Let me know if you know other usages I missed, and if you want a different naming/location for `ContextBase`.

python-spec/src/somacore/base.py
```diff
@@ -10,6 +10,7 @@
 from typing_extensions import LiteralString, Self
 
 from . import options
+from .types import ContextBase
```
thetorpedodog, line 13: minor nit: in somacore import modules, not things inside modules, so `from . import types` and use `types.ContextBase` below

python-spec/src/somacore/base.py
```diff
@@ -24,7 +25,7 @@ def open(
         uri: str,
         mode: options.OpenMode = "r",
         *,
-        context: Optional[Any] = None,
+        context: Optional[ContextBase] = None,
```
thetorpedodog, line 28: I was thinking about this a little further and it turns out there is a deeper wrinkle to this: we wouldn’t be able to override

```python

class SOMAObject(...):
  @classmethod
  def open(..., context: Optional[ContextBase] = None) -> Self: ...
```

with

```python
class ImplObject(somacore.SOMAObject):
  @classmethod
  def open(..., context: Optional[ImplContext] = None) -> Self: ...
  # not allowed, because `ImplContext` is a narrowing of `ContextBase`,
  # which would violate the Liskov principle:
  # https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

We could do a bunch of chicanery with another entry in the list of generics, so it would look something like:

```python
_CtxT = TypeVar("_CtxT", bound=ContextBase)
class SOMAObject(Generic[_CtxT]):
  def open(..., context: Optional[_CtxT] = None) -> Self: ...

# elsewhere:

class ImplObject(somacore.SOMAObject[ImplContext]):
  def open(..., context: Optional[ImplContext] = None) -> Self: ...
```

but that would also require adding a `_CtxT` to *every* SOMA object, which would be particularly unwieldy on the already highly-genericised Measurement type:

https://github.com/single-cell-data/SOMA/blob/61f6cc9e1608bd4a33110056e0530e434c2e1495/python-spec/src/somacore/measurement.py#L26-L29

so instead, I think the easier thing to do is to opt out of type-checking for this parameter when being passed in, which, while imperfect, will allow it to be overridden with type information in the implementation. So the end result looks like:

```python
class SOMAObject(metaclass=abc.ABCMeta):
  @classmethod
  def open(..., context: Optional[Any] = None) -> Self:
    ...
  # Ignore type-checking the specific type allowed in `open`

  @property
  def context(self) -> Optional[types.ContextBase]: ...
  # Returning a subclass is OK, though.
```

Then, at the implementation, we can say:

```python
class ImplObject(SOMAObject):
  @classmethod
  def open(..., context: Optional[ImplContext] = None) -> Self: ...
  # Allowed since the superclass `context` parameter type is unchecked.

  @property
  def context(self) -> Optional[ImplContext]: ...
  # Allowed since you can narrow returned types.
```

python-spec/src/somacore/ephemeral/collections.py
```diff
@@ -51,7 +52,9 @@ def open(cls, *args, **kwargs) -> NoReturn:
         )
 
     @classmethod
-    def exists(cls, uri: str, *, context: Any = None) -> Literal[False]:
+    def exists(
+        cls, uri: str, *, context: Optional[ContextBase] = None
+    ) -> Literal[False]:
```
thetorpedodog, line 57: In light of the above, this can be left as `context: Any = None` (i.e. no changes to this file at all).

python-spec/src/somacore/query/query.py
```diff
@@ -587,11 +588,16 @@ def _var_df(self) -> data.DataFrame:
 
     @property
     def _threadpool(self) -> futures.ThreadPoolExecutor:
-        """Creates a thread pool just in time."""
+        """
+        Returns the threadpool provided by the experiment's context.
+        If not available, creates a thread pool just in time."""
+        if (
+            self.experiment.context is not None
+            and self.experiment.context.threadpool is not None
```
thetorpedodog, line 596: nit: since `context` and `context.threadpool` will be nonzero if they are set at all, this can be

```python
context = self.experiment.context
if context and context.threadpool:
  return context.threadpool
```

without the need for `is not None`.

python-spec/src/somacore/types.py
```diff
@@ -75,3 +76,12 @@ def is_slice_of(__obj: object, __typ: Type[_T]) -> TypeGuard[Slice[_T]]:
         and (__obj.stop is None or isinstance(__obj.stop, __typ))
         and (__obj.step is None or isinstance(__obj.step, __typ))
     )
+
+
+class ContextBase(Protocol):
+    """A protocol for a context manager that can be used as a base class.
+    If a threadpool is specified as part of the context, it will be used by the
+    implementer. Otherwise, the implementer will use its own threadpool.
```
thetorpedodog, line 84: "it will be used by experiment queries", maybe? the implementer would be the one providing it

thetorpedodog (commented):
Thought about this a bit more and realized some annoying caveats (which exist for legitimate reasons) that would come from the original way I suggested doing this.

bkmartinjr (approved):
LGTM

PR #185, opened 2024-01-10
Prevent deadlocks in EagerIterators by making prefetch optional.

thetorpedodog:
Previously, if you provided a thread pool that was too small and an EagerIterator could not create a new preloading thread, the iterator would deadlock, since it would wait for the new thread to be created forever and not try to just do the work itself. This change instead uses preloading as an optional optimization, and if the preload has not yet been completed, computes the next value itself.


bkmartinjr (approved):

ebezzi (approved):
Looks great, thanks Paul!

PR #187, opened 2024-03-07
Add `pa.ChunkedArray` to `SparseNDCoord` type

ryan-williams:
For discussion, cf. https://github.com/single-cell-data/TileDB-SOMA/issues/1791, cc @johnkerl @bkmartinjr.

What more might we want to do here? Existing TileDB-SOMA tests pass with and without this change, see https://github.com/single-cell-data/TileDB-SOMA/pull/2234


bkmartinjr (approved):
good catch!

PR #189, opened 2024-03-20
[python] Use ` _index_factory` in `AxisIndexer` class

jp-dark:
Replace `pd.Index` with the user provided `_index_factory` when construction the cached var/obs indices.


python-spec/src/somacore/query/query.py
```diff
@@ -757,14 +757,18 @@ class AxisIndexer:
     def _obs_index(self) -> pd.Index:
         """Private. Return an index for the ``obs`` axis."""
         if self._cached_obs is None:
-            self._cached_obs = pd.Index(data=self.query.obs_joinids().to_numpy())
+            self._cached_obs = self._index_factory(
+                data=self.query.obs_joinids().to_numpy()
```
thetorpedodog, line 761: the mypy error here is because it’s expecting the argument to be positional; removing `data=` should do it.

python-spec/src/somacore/query/query.py
```diff
@@ -757,14 +757,18 @@ class AxisIndexer:
     def _obs_index(self) -> pd.Index:
```
thetorpedodog, line 757: all the appearances of `pd.Index` in this class should probably be replaced with `types.IndexLike`.

thetorpedodog (commented):
A couple minor things that I think are technically only relevant to static analysis. It is probably worth running a tiledbsoma pipeline with the `somacore` dep pointed at this change to make sure it works end-to-end.

jp-dark:
> It is probably worth running a tiledbsoma pipeline with the `somacore` dep pointed at this change to make sure it works end-to-end.

Tested this locally and all tests passed.

thetorpedodog (approved):

PR #195, opened 2024-05-09
(WIP) Add spatial support to a standard ExperimentAxisQuery

jp-dark:
This PR adds support for querying spatial data by obs/var.

Done:
* Query `obsl` dataframe

To do:
* Query `scene_id` from `ExperimentAxisQuery`
* Query `obssm` dataframe
* Query `varssm` dataframe
* Query `varl/[measure]` dataframe
* Query images

### DIFF
--- a/python-spec/src/somacore/__init__.py
+++ b/python-spec/src/somacore/__init__.py
@@ -16,6 +16,9 @@ import pyarrow_hotfix
 
 from .base import SOMAObject
 from .collection import Collection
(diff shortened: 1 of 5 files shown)


prathapsridharan:
@jp-dark @pablo-gar - Per the new PR below, `obsl` is now a dataframe and not a collection right? So this should be modified?

https://github.com/single-cell-data/SOMA/pull/196

PR #196, opened 2024-05-14
Proposed update to the spatial additions

jp-dark:
This PR provides a proposed restructuring to the spatial components.

### DIFF
--- a/python-spec/src/somacore/ephemeral/collections.py
+++ b/python-spec/src/somacore/ephemeral/collections.py
@@ -122,7 +122,7 @@ _BasicAbstractMeasurement = measurement.Measurement[
 """The loosest possible constraint of the abstract Measurement type."""
 
 _BasicAbstractScene = scene.Scene[
-    collection.Collection[Union[data.DataFrame, data.NDArray]], base.SOMAObject
+    data.DataFrame, collection.Collection[data.NDArray], base.SOMAObject
 ]
 """The loosest possible constraint of the abstract Scene type."""
 
--- a/python-spec/src/somacore/ephemeral/collections.py
+++ b/python-spec/src/somacore/ephemeral/collections.py
@@ -1,6 +1,6 @@
 from typing import Any, Dict, Iterator, NoReturn, Optional, TypeVar
 
-from typing_extensions import Literal, Self, Union
+from typing_extensions import Literal, Self
 
 from .. import base
 from .. import collection
@@ -148,7 +148,7 @@ class Experiment(  # type: ignore[misc]  # __eq__ false positive
     experiment.Experiment[
         data.DataFrame,
         collection.Collection[_BasicAbstractMeasurement],
-        collection.Collection[Union[data.DataFrame, _BasicAbstractScene]],
+        collection.Collection[_BasicAbstractScene],
         base.SOMAObject,
     ],
 ):
--- a/python-spec/src/somacore/experiment.py
+++ b/python-spec/src/somacore/experiment.py
@@ -66,10 +66,15 @@ class Experiment(
     spatial = _mixin.item[_SceneColl]()  # TODO: Discuss the name of this element.
     """A collection of named spatial scenes."""
 
-    obs_scene = _mixin.item[_DF]()  # TODO: Rename this element.
+    # TODO: Discuss the name of this element.
+    # TODO: Discuss the best location for this element. Should it be moved
+    # inside the spatial collection?
+    # TODO: Use DF or new array type for int/string indexing?
+    obs_spatial = _mixin.item[_DF]()
     """A dataframe mapping obs_id to scene names.
 
-    This provides a join table for the obs and scene ids.
+    This provides a join table for the obs ``soma_joinid`` and the ``scene_ids``
+    used in the ``spatial`` collection.
     """
 
     def axis_query(
--- a/python-spec/src/somacore/experiment.py
+++ b/python-spec/src/somacore/experiment.py
@@ -1,4 +1,4 @@
-from typing import Generic, Optional, TypeVar, Union
+from typing import Generic, Optional, TypeVar
 
 from typing_extensions import Final, Self
 
@@ -14,9 +14,7 @@ _DF = TypeVar("_DF", bound=data.DataFrame)
 """An implementation of a DataFrame."""
 _MeasColl = TypeVar("_MeasColl", bound=collection.Collection[measurement.Measurement])
 """An implementation of a collection of Measurements."""
-_SceneColl = TypeVar(
-    "_SceneColl", bound=collection.Collection[Union[data.DataFrame, scene.Scene]]
-)
+_SceneColl = TypeVar("_SceneColl", bound=collection.Collection[scene.Scene])
 """An implemenation of a collection of spatial data."""
 _RootSO = TypeVar("_RootSO", bound=base.SOMAObject)
 """The root SOMA object type of the implementation."""
@@ -65,8 +63,14 @@ class Experiment(
     ms = _mixin.item[_MeasColl]()
     """A collection of named measurements."""
 
-    spatial = _mixin.item[_SceneColl]()
-    """TODO Add docs for spatial collecion."""
+    spatial = _mixin.item[_SceneColl]()  # TODO: Discuss the name of this element.
+    """A collection of named spatial scenes."""
+
+    obs_scene = _mixin.item[_DF]()  # TODO: Rename this element.
+    """A dataframe mapping obs_id to scene names.
+
+    This provides a join table for the obs and scene ids.
+    """
 
     def axis_query(
         self,
--- a/python-spec/src/somacore/scene.py
+++ b/python-spec/src/somacore/scene.py
@@ -1,6 +1,6 @@
 """Implementation of the SOMA scene collection for spatial data"""
 
-from typing import Generic, TypeVar, Union
+from typing import Generic, TypeVar
 
 from typing_extensions import Final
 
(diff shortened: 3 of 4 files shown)


jp-dark:
Summary of proposed changes:

Top level changes:

* The `spatial` collection now only contains the scene. All catalog metadata gets bumped to the Experiment level.
* Add new dataframe (name TBD) to map between the `obs_id` and `scene_id`.
  - Decision: Should `scene_id` be an integer `soma_joinid` or the names of the Scenes stored in `spatial`?

Scene changes:

* Add `img` collection for images.
* Change `obsl` and `varl` to store location data only.
* Add `obsm` and `varsm` (names TBD) for additional obs/var dataframes/point clouds.

python-spec/src/somacore/experiment.py
```diff
@@ -65,8 +63,14 @@ class Experiment(
     ms = _mixin.item[_MeasColl]()
     """A collection of named measurements."""
 
-    spatial = _mixin.item[_SceneColl]()
-    """TODO Add docs for spatial collecion."""
+    spatial = _mixin.item[_SceneColl]()  # TODO: Discuss the name of this element.
+    """A collection of named spatial scenes."""
+
+    obs_scene = _mixin.item[_DF]()  # TODO: Rename this element.
```
jp-dark, line 69: Should we also have a join table between var/scene?

prathapsridharan: @jp-dark - 

1. Am I correct in understanding that this is a dataframe that captures the _many-to-many_ relationship between observations and scenes?

That is, this is a dataframe that maps the primary key in `obs` (`obs_id`) to the primary key for a scene which is of type `string` that can be used to index the `spatial` collection in `Experiment`?

2. Will this dataframe store anything more than just a mapping of primary keys of `obs` to primary keys of `scenes`?

prathapsridharan (changes requested):

prathapsridharan:
> * The `spatial` collection now only contains the scene. All catalog metadata gets bumped to the Experiment level.

@jp-dark - 

1. I might have missed it but where is the catalog metadata stored in at the Experiment level? I only see `obs` and `obs_scene` dataframes in `Experiment`.
2. By _"catalog metadata"_ are referring to metadata about _scenes_ or something else?

>   * Decision: Should `scene_id` be an integer `soma_joinid` or the names of the Scenes stored in `spatial`?

@jp-dark @pablo-gar - Correct me if I am wrong but my understanding of the a `Collection` is that it is a _string keyed map_. If `spatial` is such a collection and if it is to be keyed by a `scene_id`, then wouldn't the `scene_id` necessarily have to be a `string`?

jp-dark:
>  > * The `spatial` collection now only contains the scene. All catalog metadata gets bumped to the Experiment level.
>
>     1. I might have missed it but where is the catalog metadata stored in at the Experiment level? I only see `obs` and `obs_scene` dataframes in `Experiment`.
> 
>     2. By _"catalog metadata"_ are referring to metadata about _scenes_ or something else?

"Catalog metadata" was a poor descriptor on my part. I'm talking about any dataframes/arrays that describe the scenes and/or the relationships between scenes are other pieces of the Experiement (right now just `obs_scene` but we might want to add more).

> > * Decision: Should `scene_id` be an integer `soma_joinid` or the names of the Scenes stored in `spatial`?
> 
> @jp-dark @pablo-gar - Correct me if I am wrong but my understanding of the a `Collection` is that it is a _string keyed map_. If `spatial` is such a collection and if it is to be keyed by a `scene_id`, then wouldn't the `scene_id` necessarily have to be a `string`?

As is, yes**. If we key on a `soma_joinid`, we would need to add a mapping from the scene name to the join ID.

** Groups in TileDB are actually primarily keyed by index with an optional string name, but I believe the SOMA implementation treats collections as string keyed maps.

python-spec/src/somacore/experiment.py
prathapsridharan: @jp-dark - I am also inclined to agree with using a `string` for `scene_id` and removing the indirection. We can always add it later if we think it is useful but right now it seem more straightforward given that Collections are indexed by a string key and if metadata about a scene needs to be looked up in a dataframe, it seems simpler to use that same string key as the primary key in such a "metadata dataframe"

python-spec/src/somacore/scene.py
```diff
@@ -36,17 +38,27 @@ class Scene(
...
+    img = _mixin.item[_ImageColl]()
+    """A collection of imagery backing the spatial data"""
 
-    varl = _mixin.item[collection.Collection[_SpatialColl]]()
-    """A collection of collections of spatial data defined on a measurement variable"""
+    obsl = _mixin.item[_SpatialDF]()
+    """A dataframe of the obs locations"""
+
+    varl = _mixin.item[collection.Collection[_SpatialDF]]()
+    """A collection of dataframes of the var locations"""
```
prathapsridharan, line 58: @jp-dark - `varl` Collection - Is this indexed by `scene_id` and the value is a dataframe that has: `obs_id`, `var_id`, and some geometry columns?

jp-dark: > varl Collection - Is this indexed by scene_id and the value is a dataframe that has: obs_id, var_id, and some geometry columns?

No, the collection is indexed by the measurement name. The value is a dataframe that has `var_id`, some geometry columns, and possibly other accessory data.

python-spec/src/somacore/experiment.py
```diff
@@ -65,8 +63,14 @@ class Experiment(
     ms = _mixin.item[_MeasColl]()
     """A collection of named measurements."""
 
-    spatial = _mixin.item[_SceneColl]()
-    """TODO Add docs for spatial collecion."""
+    spatial = _mixin.item[_SceneColl]()  # TODO: Discuss the name of this element.
```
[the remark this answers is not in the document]
jp-dark: > 1. In spatial soma, is `spatial` a collection used to access a Scene object?

Yes. More precisely it is the `Collection` that the scene objects are contained inside of. 

> 2. If so, what is the key used to access a `Scene` in container? experiment.spatial[???] - Here the key is of type string right? Maybe a concrete example would help.

The `Scene` name is arbitrary and up to the user, just like for a `Measurement`. For example, for a collection of Visium datasets, it would make sense to use the `dataset_id` for the `Scene` name.

python-spec/src/somacore/ephemeral/collections.py
```diff
@@ -122,7 +122,7 @@ class Collection(  # type: ignore[misc]  # __eq__ false positive
 """The loosest possible constraint of the abstract Measurement type."""
 
 _BasicAbstractScene = scene.Scene[
-    collection.Collection[Union[data.DataFrame, data.NDArray]], base.SOMAObject
+    data.DataFrame, collection.Collection[data.NDArray], base.SOMAObject
```
pablo-gar, line 125: Not sure I understand this. Are we saying a Scene _could_ be a data frame OR a collection?

python-spec/src/somacore/experiment.py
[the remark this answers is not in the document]
pablo-gar: @prathapsridharan and @jp-dark We discussed in our last meeting that and integer may be better, and in my view we settled on using integers for `scene_id`

python-spec/src/somacore/experiment.py
pablo-gar: I'll defe to you guys, but in our last conversation we had settled on integers. Maybe we should revisit the conversation?

> Should we also have a join table between var/scene?

I think so, it may not be very useful for Visium/Slide-seq but likely useful for Xenium

python-spec/src/somacore/scene.py
```diff
@@ -9,19 +9,21 @@
 from . import collection
 from . import data
 
-_DF = TypeVar("_DF", bound=data.DataFrame)
-"""A particular implementation of DataFrame."""
-_SpatialColl = TypeVar(
-    "_SpatialColl", bound=collection.Collection[Union[data.DataFrame, data.NDArray]]
-)
+_SpatialDF = TypeVar(
+    "_SpatialDF", bound=data.DataFrame
```
pablo-gar, line 13: @jp-dark, are we settled on that we will always need exactly one spatial DF per scene? or are there any cases in which we need to follow the same logic as _ImageColl being a collection?

I can't think of a case for multiple spatialDFs, so I'm ok with this. But it may be worth checking with @aaronwolen

pablo-gar (changes requested):
Only minor comments. Let me know when another round of review is needed

python-spec/src/somacore/ephemeral/collections.py
jp-dark: There is a comment in the `scene.Scene` class definition:

```
    # This class is implemented as a mixin to be used with SOMA classes.
    # For example, a SOMA implementation would look like this:
    #
    #     # This type-ignore comment will always be needed due to limitations
    #     # of type annotations; it is (currently) expected.
    #     class Scene(  # type: ignore[type-var]
    #         ImplBaseCollection[ImplSOMAObject],
    #         somacore.Scene[
    #             Union[ImplGeometryDataFrame, ImplPointCloud], # _SpatialDF
    #             ImplImageCollection,                          # _ImageColl
    #             ImplSOMAObject,                               # _RootSO
    #         ],
    #     ):
    #         ...
```

The `_BasicAbstractScene` is just defining the inner `somacore.Scene` piece.

python-spec/src/somacore/scene.py
jp-dark: For the TileDB implementation, you may want to split up a spatial dataframe into multiple TileDB arrays. For example, the geospatial team does that for large point clouds using octrees for look-up. However, I think this type would still be a GeometryDataFrame or PointCloud type with extra additional functions for retrieving subsets of the data.

That being said, we should explore this question in more detail.

python-spec/src/somacore/experiment.py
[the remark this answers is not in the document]
pablo-gar: After a discussion we settled on strings

prathapsridharan:
@jp-dark (cc: @pablo-gar) - For the `obsl`, `varl`, and `obssm` and `varssm` some more detailed comments about their meaning would be helpful. If possible, an illustrative example would not be bad either - For instance, `obsl` contiains columns: `obs_soma_joinid: int`, `geometry: SomaGeometry`, etc. 

Especially since we will inevitably be changing things, these comments will be helpful for the future.

python-spec/src/somacore/experiment.py
jp-dark: After discussion we decided to use strings for `scene_id` for the time being.

prathapsridharan (approved):
Thank you @jp-dark for patiently addressing the questions in this PR and moving it along.

PR #199, opened 2024-06-14
Add first pass at obs/var to scene implementation

jp-dark:
This is a prototype implementation for getting scene IDs from the obs or var index on an experiment query axis for testing purposes.

Issue: [#2592](https://github.com/single-cell-data/TileDB-SOMA/issues/2592)


jp-dark:
[sc-46720]

prathapsridharan (approved):

PR #200, opened 2024-06-14
Python spec for geometry dataframe

jp-dark:
Changes:
* Add a new GeometryDataFrame class
* Adds shapely as a somacore Python dependency.

WIP:
* How we want to handle the two-dimensional geometry index dimension for both schema specification needs further investigation.

Issues: Closes https://github.com/single-cell-data/TileDB-SOMA/issues/2677


pablo-gar (approved):
I reviewed the naming of the introduced object and methods. I defer to others on the API review.

PR #201, opened 2024-06-17
Type-check Pandas code and require newer pyarrow.

thetorpedodog:
- Pandas now has canonical type stubs, which mypy can find and install.
- We no longer support py3.7, so we don't need the pyarrow hotfix.
- Type annotation fix related to installing Pandas stubs.


johnkerl:
> We no longer support py3.7, so we don't need the pyarrow hotfix.

@thetorpedodog this is a longer story -- we don't need the hotfix _as long as_ we've got pyarrow >= 14

But here is the current status on that, which is not done yet:

* https://github.com/single-cell-data/TileDB-SOMA/pull/2692
* https://github.com/apache/arrow/issues/42154#issuecomment-2173608139

This is confusing, so to be extra-clear here:

* On MacOS, there is a fatal error unless pyarrow < 13
* We've lived that way for a while
* Trying to have TileDB-SOMA support Python 3.12 (https://github.com/single-cell-data/TileDB-SOMA/pull/2692) -- is blocked since pyarrow < 13 doesn't exist with Python 3.12
* This has also, separately, impacted the attempted TileDB-SOMA 1.12 release https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/171 but in a way which has been worked around for now.

TL;DR -- we need `pyarrow-hotfix` for now. And I'll deal with this, once all the above issues are resolved, on https://github.com/single-cell-data/TileDB-SOMA/issues/1926

johnkerl (changes requested):
> We no longer support py3.7, so we don't need the pyarrow hotfix.

@thetorpedodog this is a longer story -- we don't need the hotfix _as long as_ we've got pyarrow >= 14

But here is the current status on that, which is not done yet:

* https://github.com/single-cell-data/TileDB-SOMA/pull/2692
* https://github.com/apache/arrow/issues/42154#issuecomment-2173608139

This is confusing, so to be extra-clear here:

* On MacOS, there is a fatal error unless pyarrow < 13
* We've lived that way for a while
* Trying to have TileDB-SOMA support Python 3.12 (https://github.com/single-cell-data/TileDB-SOMA/pull/2692) -- is blocked since pyarrow < 13 doesn't exist with Python 3.12
* This has also, separately, impacted the attempted TileDB-SOMA 1.12 release https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/171 but in a way which has been worked around for now.

TL;DR -- we need `pyarrow-hotfix` for now. And I'll deal with this, once all the above issues are resolved, on https://github.com/single-cell-data/TileDB-SOMA/issues/1926

pyproject.toml
```diff
@@ -16,12 +16,9 @@ dependencies = [
   "numba",
   "numpy>=1.21",
   "pandas",
-  "pyarrow",
-  # TODO: once we no longer support Python 3.7, remove this and pin to pyarrow >= 14.0.1
```
johnkerl, line 20: Following on from my review comment: let's keep `pyarrow-hotfix` _and_ update the reason-why-it-is-needed comment (which has gone stale)

thetorpedodog:
To do this, I (after some trial and error):

 1. Removed all the entries from the `[[tool.mypy.overrides]] module = [...]` setting, so that mypy would no longer ignore any modules
 2. Ran `update-requirements-txt` to rebuild the requirements files.
 3. Looked at the mypy output from `update-requirements-txt` and re-enabled ignoring modules that had no type stubs (and also fixed any issues that arose).

pyproject.toml
thetorpedodog: pyarrow-hotfix has been restored.

johnkerl (approved):

PR #204, opened 2024-07-24
Drop unused categories in `ExperimentAxisQuery.to_anndata`

nguyenv:
**Issue:**
https://github.com/single-cell-data/TileDB-SOMA/issues/2765

**Changes:**
`ExperimentAxisQuery.to_anndata` iterates through each column and drops unused categories.

**Notes for Reviewer:**
This change was originally proposed in https://github.com/single-cell-data/TileDB-SOMA/pull/2811 but was suggested somacore was a more appropriate location for it.

### DIFF
--- a/python-spec/src/somacore/query/query.py
+++ b/python-spec/src/somacore/query/query.py
@@ -277,6 +277,7 @@ class ExperimentAxisQuery(Generic[_Exp]):
         obsp_layers: Sequence[str] = (),
         varm_layers: Sequence[str] = (),
         varp_layers: Sequence[str] = (),
+        drop_levels: bool = False,
     ) -> anndata.AnnData:
         """
         Executes the query and return result as an ``AnnData`` in-memory object.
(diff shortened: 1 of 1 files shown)


johnkerl (changes requested):
From the discussion here

https://github.com/single-cell-data/TileDB-SOMA/issues/2765#issuecomment-2248061969

we need a new flag on `to_anndata`:

* optional
* default if omitted: current behavior
* if supplied and `True`: do what this PR does
* name: `droplevels`? Or something else? Let's chat with @pablo-gar @mojaveazure @eddelbuettel

python-spec/src/somacore/query/query.py
```diff
@@ -298,7 +299,7 @@ def to_anndata(
```
johnkerl, line 299: Let's add a description of what `drop_levels` does ...

johnkerl: @pablo-gar @eddelbuettel @mojaveazure are you okay with `drop_levels`?

johnkerl: (Presuming we'll want a same-ish name in Python & R both ...)

mojaveazure: Seurat uses `drop` to match R, but I'm fine with `drop_levels` in SOMA

python-spec/src/somacore/query/query.py
```diff
@@ -295,10 +296,13 @@ def to_anndata(
                 Additional varm layers to read and return in the varm slot.
             varp_layers:
                 Additional varp layers to read and return in the varp slot.
+            drop_levels:
+                Indicate whether unused categories on axis frames should be
+                dropped. By default, False, the categories are not dropped.
```
johnkerl, line 301:
```suggestion
                dropped. By default, False, the categories which are present in the SOMA Experiment
                and not present in the query output are not dropped.
```

johnkerl (approved):
🚢 with a suggested docstring update

python-spec/src/somacore/query/query.py
pablo-gar: Sounds good to me!

PR #206, opened 2024-07-26
Prototype image collection

jp-dark:
This PR adds a prototype image collection. The following decisions should be reviewed:

* Proposed API
* Naming choices

### DIFF
--- a/python-spec/src/somacore/ephemeral/collections.py
+++ b/python-spec/src/somacore/ephemeral/collections.py
@@ -132,9 +132,7 @@ _BasicAbstractMeasurement = measurement.Measurement[
 ]
 """The loosest possible constraint of the abstract Measurement type."""
 
-_BasicAbstractScene = scene.Scene[
-    data.DataFrame, collection.Collection[data.NDArray], base.SOMAObject
-]
+_BasicAbstractScene = scene.Scene[data.DataFrame, images.Image2D, base.SOMAObject]
 """The loosest possible constraint of the abstract Scene type."""
 
 
--- a/python-spec/src/somacore/ephemeral/collections.py
+++ b/python-spec/src/somacore/ephemeral/collections.py
@@ -1,5 +1,14 @@
-from typing import Any, Dict, Iterator, MutableMapping, NoReturn, Optional, TypeVar
+from typing import (
+    Any,
+    Dict,
+    Iterator,
+    MutableMapping,
+    NoReturn,
+    Optional,
+    TypeVar,
+)
 
+import pyarrow as pa
 from typing_extensions import Literal, Self
 
 from .. import base
@@ -7,6 +16,7 @@ from .. import collection
 from .. import coordinates
 from .. import data
 from .. import experiment
+from .. import images
 from .. import measurement
 from .. import options
 from .. import scene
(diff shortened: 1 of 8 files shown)


python-spec/src/somacore/ephemeral/collections.py
```diff
@@ -154,6 +162,32 @@ def transformations(self) -> MutableMapping[str, coordinates.CoordinateTransform
         raise NotImplementedError()
 
 
+class Image2D(  # type: ignore[misc]   # __eq__ false positive
+    BaseCollection[base.SOMAObject], images.Image2D[data.DenseNDArray, base.SOMAObject]
+):
```
jp-dark, line 167: I created an image collection for 2D images only. This is slightly different from early high-level discussions. I am not adding a 3D image collection now, but we can add one in the future.

python-spec/src/somacore/ephemeral/collections.py
```diff
@@ -154,6 +162,32 @@ def transformations(self) -> MutableMapping[str, coordinates.CoordinateTransform
...
+    def read_level(
+        self,
+        level: int,
+        coords: options.DenseNDCoords = (),
+        *,
+        transform: Optional[coordinates.CoordinateTransform] = None,
+        result_order: options.ResultOrderStr = options.ResultOrder.AUTO,
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> pa.Tensor:
+        raise NotImplementedError()
```
jp-dark, line 188: This function will be modified after I create a return type for spatial reads. The new return type will include coordinate space information and the ability to lazily apply coordinate transforms.

python-spec/src/somacore/images.py
```diff
@@ -0,0 +1,81 @@
...
+_RootSO = TypeVar("_RootSO", bound=base.SOMAObject)
+"""The root SomaObject type of the implementation."""
+
+_RO_AUTO = options.ResultOrder.AUTO
+
+
+class Image2D(
+    collection.BaseCollection[_RootSO],
+    Generic[_DenseND, _RootSO],
+):
```
jp-dark, line 26: I created an image collection for 2D images only. This is slightly different from early high-level discussions. I am not adding a 3D image collection now, but we can add one in the future.

johnkerl (approved):

python-spec/src/somacore/images.py
```diff
@@ -0,0 +1,93 @@
...
+
+    __slots__ = ()
+    soma_type: Final = "SOMAImage2D"  # type: ignore[misc]
+
+    class LevelProperties(Protocol):
+        """Class requirements for Image2D level properties."""
+
+        @property
+        def axes(self) -> Tuple[str, ...]:
+            """Axis order for the underlying data."""
```
ivirshup, line 51: Shouldn't the axes be the same for every level of the image?

python-spec/src/somacore/ephemeral/collections.py
```diff
@@ -154,6 +164,43 @@ def transformations(self) -> MutableMapping[str, coordinates.CoordinateTransform
         raise NotImplementedError()
 
 
+class Image2D(  # type: ignore[misc]   # __eq__ false positive
```
ivirshup, line 167: Since this are explicitly multiscale, I think it would make sense if that was clear in the name. Maybe `MultiscaleImage2D`? I think there's a version of the spec where these were called `ImageCollection`?

python-spec/src/somacore/images.py
pablo-gar: I think we should name this `Image2DCollection` to make it clear it is one, otherwise it can be confused to and actual image array. It also follow the general naming structure of other SOMA objects, e.g. Sparse**Array** and Dense**Array**

python-spec/src/somacore/images.py
```diff
@@ -0,0 +1,93 @@
...
+    @abc.abstractmethod
+    def read_level(
+        self,
+        level: int,
+        coords: options.DenseNDCoords = (),
+        *,
+        transform: Optional[coordinates.CoordinateTransform] = None,
+        result_order: options.ResultOrderStr = _RO_AUTO,
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> pa.Tensor:
```
pablo-gar, line 91: probably for my understanding, why are we reading into a `pa.Tensor`? should we not do so into an iterator?

python-spec/src/somacore/query/query.py
```diff
@@ -871,8 +871,7 @@ def obs(self) -> data.DataFrame: ...
     def context(self) -> Optional[base_types.ContextBase]: ...
 
     @property
-    def obs_scene(self) -> data.DataFrame:
-        ...
+    def obs_scene(self) -> data.DataFrame: ...
```
pablo-gar, line 874: Maybe introduce the latest names we discussed:

`obs_spatial_presence`

pablo-gar (changes requested):

python-spec/src/somacore/images.py
pablo-gar: Oh yeah right. Hm I think this defeats our goals for out-of-core processing, for images this is going to be important since we don't want to load GBs of data into memory.

I think for when we land the 1.0 of spatial we should either:

1. Enable a blockwise-like iterator for dense arrays.
2. Use sparse arrays until dense arrays are truly out-of-core.

@jp-dark I'm not sure what would be best for now in these early iterations

PR #218, opened 2024-09-12
Add classes for coordinate spaces and transforms

jp-dark:
New classes:
* Axis: simple data class storing a name, optional units, and optional scale
* CoordinateSpace: used to define coordinate spaces for spatial data
* CoordinateTransform: abstract base class for coordinate transforms
* AffineTransform: implementation for a generic affine coordinate transform
* ScaleTransform: implementation for a coordinate transform that scales axes (may be isotropic/uniform)
* IdentityTransform: implementation for an identity transform that only applies a name change to axes

### DIFF
--- a/python-spec/src/somacore/__init__.py
+++ b/python-spec/src/somacore/__init__.py
@@ -22,6 +22,7 @@ from .coordinates import CoordinateSpace
 from .coordinates import CoordinateTransform
 from .coordinates import IdentityTransform
 from .coordinates import ScaleTransform
+from .coordinates import UniformScaleTransform
 from .data import DataFrame
 from .data import DenseNDArray
 from .data import NDArray
@@ -72,5 +73,6 @@ __all__ = (
     "CoordinateTransform",
     "AffineTransform",
     "ScaleTransform",
+    "UniformScaleTransform",
     "IdentityTransform",
 )
--- a/python-spec/src/somacore/__init__.py
+++ b/python-spec/src/somacore/__init__.py
@@ -16,6 +16,12 @@ import pyarrow_hotfix  # noqa: F401
 
 from .base import SOMAObject
 from .collection import Collection
+from .coordinates import AffineTransform
+from .coordinates import Axis
+from .coordinates import CoordinateSpace
+from .coordinates import CoordinateTransform
+from .coordinates import IdentityTransform
+from .coordinates import ScaleTransform
 from .data import DataFrame
 from .data import DenseNDArray
 from .data import NDArray
@@ -61,4 +67,10 @@ __all__ = (
     "AxisQuery",
     "ExperimentAxisQuery",
     "ContextBase",
+    "Axis",
+    "CoordinateSpace",
+    "CoordinateTransform",
+    "AffineTransform",
+    "ScaleTransform",
+    "IdentityTransform",
 )
--- a/python-spec/src/somacore/coordinates.py
+++ b/python-spec/src/somacore/coordinates.py
@@ -195,8 +195,6 @@ class AffineTransform(CoordinateTransform):
                 f"Matrix multiply is not implemented with type {type(other)!r}."
             )
         self._check_matmul_inner_axes(other)
-        if self.input_axes != other.output_axes:
-            raise ValueError("Axis mismatch between transformations.")
         if isinstance(other, IdentityTransform):
             return AffineTransform(other.input_axes, self.output_axes, self._matrix)
         if isinstance(other, AffineTransform):
--- a/python-spec/src/somacore/coordinates.py
+++ b/python-spec/src/somacore/coordinates.py
@@ -1,10 +1,10 @@
 """Definitions of types related to coordinate systems."""
 
 import abc
-import attrs
 import collections.abc
 from typing import Any, Optional, Sequence, Tuple, Union
 
+import attrs
 import numpy as np
 import numpy.typing as npt
 
--- a/python-spec/src/somacore/coordinates.py
+++ b/python-spec/src/somacore/coordinates.py
@@ -15,15 +15,15 @@ class Axis(metaclass=abc.ABCMeta):
 
     Args:
         name: Name of the axis.
-        unit: Optional string units. Defaults to ``None``.
-        scale: Optional scale for units. Defaults to ``None``.
+        unit_name: Optional string name for the units. Defaults to ``None``.
+        unit_scale: Optional scale for units. Defaults to ``None``.
 
     Lifecycle: experimental
     """
 
     name: str
-    units: Optional[str] = None
-    scale: Optional[np.float64] = None
+    unit_name: Optional[str] = None
+    unit_scale: Optional[np.float64] = None
 
 
 class CoordinateSpace(collections.abc.Sequence):
@@ -77,6 +77,8 @@ class CoordinateTransform(metaclass=abc.ABCMeta):
     Args:
         input_axes: The names of the axes for the input coordinate space.
         output_axes: The names of the axes for the output coordinate space.
+
+    CoordinateTransform classes are composable using the ``@`` (__matmul__) operator.
 
     Lifecycle: experimental
     """
--- a/python-spec/src/somacore/coordinates.py
+++ b/python-spec/src/somacore/coordinates.py
@@ -1,0 +1,392 @@
+"""Definitions of types related to coordinate systems."""
+
+import abc
+import collections.abc
+from dataclasses import dataclass
+from typing import Any, Optional, Sequence, Tuple, Union
+
+import numpy as np
+import numpy.typing as npt
+
+
+@dataclass
+class Axis(metaclass=abc.ABCMeta):
+    """A description of an axis of a coordinate system
+
+    Args:
+        name: Name of the axis.
+        unit: Optional string units. Defaults to ``None``.
+        scale: Optional scale for units. Defaults to ``None``.
+
+    Lifecycle: experimental
+    """
+
+    name: str
+    units: Optional[str] = None
+    scale: Optional[np.float64] = None
+
+
+class CoordinateSpace(collections.abc.Sequence):
+    """A coordinate system for spatial data."""
+
+    def __init__(self, axes: Sequence[Axis]):
+        self._axes = tuple(axes)
+        if len(self._axes) == 0:
+            raise ValueError("Coordinate space must have at least one axis.")
+        if len(set(axis.name for axis in self._axes)) != len(axes):
+            raise ValueError("The names for the axes must be unique.")
+
+    def __len__(self) -> int:
+        return len(self._axes)
+
+    def __getitem__(self, index: int) -> Axis:  # type: ignore[override]
+        return self._axes[index]
+
+    def __repr__(self) -> str:
+        output = f"<{type(self).__name__}\n"
+        for axis in self._axes:
+            output += f"  {axis},\n"
+        return output + ">"
+
+    @property
+    def axes(self) -> Tuple[Axis, ...]:
+        """TODO: Add docstring"""
+        return self._axes
+
+    @property
+    def axis_names(self) -> Tuple[str, ...]:
+        """TODO: Add docstring"""
+        return tuple(axis.name for axis in self._axes)
+
+    def rank(self) -> int:
+        """TODO: Add docstring"""
+        return len(self)
+
+
+class CoordinateTransform(metaclass=abc.ABCMeta):
+    """TODO: Add docstring"""
+
+    def __init__(
+        self,
+        input_axes: Union[str, Sequence[str]],
+        output_axes: Union[str, Sequence[str]],
+    ):
+        self._input_axes = (
+            (input_axes,) if isinstance(input_axes, str) else tuple(input_axes)
+        )
+        self._output_axes = (
+            (output_axes,) if isinstance(output_axes, str) else tuple(output_axes)
+        )
+
+    @abc.abstractmethod
+    def __matmul__(self, other: Any) -> "CoordinateTransform":
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def __rmatmul__(self, other: Any) -> "CoordinateTransform":
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def inverse_transform(self) -> "CoordinateTransform":
+        raise NotImplementedError()
+
+    @property
+    def input_axes(self) -> Tuple[str, ...]:
+        """TODO: Add docstring"""
+        return self._input_axes
+
+    @property
+    def input_rank(self) -> int:
+        """TODO: Add docstring"""
+        return len(self._input_axes)
+
+    @property
+    def output_axes(self) -> Tuple[str, ...]:
+        """TODO: Add docstring"""
+        return self._output_axes
+
+    @property
+    def output_rank(self) -> int:
+        """TODO: Add docstring"""
+        return len(self._output_axes)
+
+
+class AffineTransform(CoordinateTransform):
+    """TODO: Add docstring"""
+
+    def __init__(
+        self,
+        input_axes: Union[str, Sequence[str]],
+        output_axes: Union[str, Sequence[str]],
+        matrix: npt.ArrayLike,
+    ):
+        super().__init__(input_axes, output_axes)
+
+        # Check the rank of the input/output axes match.
+        if self.input_rank != self.output_rank:
+            raise ValueError(
+                "The input axes and output axes must be the same length for an "
+                "affine transform."
+            )
+        rank = self.input_rank
+
+        # Create and validate the augmented matrix.
+        self._matrix: npt.NDArray[np.float64] = np.array(matrix, dtype=np.float64)
+        if self._matrix.shape == (rank + 1, rank + 1):
+            if not (
+                self._matrix[-1, -1] == 1.0
+                and np.array_equal(self._matrix[-1, :-1], np.zeros((rank,)))
+            ):
+                raise ValueError(
+                    f"Input matrix {self._matrix} has augmented matrix shape, but is not a valid "
+                    f"augmented matrix."
+                )
+        elif self._matrix.shape == (rank, rank + 1):
+            self._matrix = np.vstack(
+                (
+                    self._matrix,
+                    np.hstack((np.zeros((rank,)), np.array([1]))),
+                )
+            )
+        else:
+            raise ValueError(
+                f"Unexpected shape {self._matrix.shape} for the input affine matrix."
+            )
+
+    def __matmul__(self, other: Any) -> CoordinateTransform:
+        if isinstance(other, CoordinateTransform):
+            if self.input_axes != other.output_axes:
+                raise ValueError("Axis mismatch between transformations.")
+            if isinstance(other, IdentityTransform):
+                return AffineTransform(other.input_axes, self.output_axes, self._matrix)
+            if isinstance(other, AffineTransform):
+                return AffineTransform(
+                    other.input_axes,
+                    self.output_axes,
+                    self.augmented_matrix @ other.augmented_matrix,
+                )
+        if isinstance(other, np.ndarray):
+            raise NotImplementedError(
+                "Support for multiplying by numpy arrays is not yet implemented."
+            )
+        raise TypeError(
+            f"Cannot multiply a CoordinateTransform by type {type(other)!r}."
+        )
+
+    def __rmatmul__(self, other: Any) -> CoordinateTransform:
+        if isinstance(other, CoordinateTransform):
+            if other.input_axes != self.output_axes:
+                raise ValueError("Axis mismatch between transformations.")
+            if isinstance(other, IdentityTransform):
+                return AffineTransform(
+                    self.input_axes, other.output_axes, self.augmented_matrix
+                )
+            if isinstance(other, AffineTransform):
+                return AffineTransform(
+                    self.input_axes,
+                    other.output_axes,
+                    other.augmented_matrix @ self.augmented_matrix,
+                )
+        if isinstance(other, np.ndarray):
+            raise NotImplementedError(
+                "Support for multiplying by numpy arrays is not yet implemented."
+            )
+        raise TypeError(
+            f"Cannot multiply a CoordinateTransform by type {type(other)!r}."
+        )
+
+    @property
+    def augmented_matrix(self) -> npt.NDArray[np.float64]:
+        """Returns the augmented affine matrix for the transformation."""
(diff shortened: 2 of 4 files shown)


python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,460 @@
...
+        return len(self._output_axes)
+
+
+class AffineTransform(CoordinateTransform):
+    """An affine coordinate trasformation from one coordinate space to another.
+
+    An affine transform is a combination of a linear transformation and a translation.
+    For a vector x it can be written as
+
+        y = Ax + b
```
johnkerl, line 130: This is a little confusing. In this line you're showing `A` as non-augmented but below you are describing it in augmented form. Maybe just a sentence here saying the "+ b" part can be folded into the matrix using the standard augmented-matrix trick ...

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,460 @@
...
+
+    where A is a matrix and b is a vector.
+
+
+    Args:
+        input_axes: The names of the axes for the input coordinate space.
+        output_axes: The names of the axes for the output coordinate space.
+        matrix: Matrix for that represetnts the transformation. Can be provided as
```
johnkerl, line 138:
```suggestion
        matrix: Matrix (perhaps augmented) that represents the affine transformation. Can be provided as
```

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,460 @@
...
+                return ScaleTransform(
+                    other.input_axes,
+                    self.output_axes,
+                    self.scale_factors * other.scale_factors,
+                )
+            if isinstance(other, AffineTransform):
+                return AffineTransform(
+                    other.input_axes,
+                    self.output_axes,
+                    self.augmented_matrix @ other.augmented_matrix,
```
johnkerl, line 304: Maybe a comment for the non-experts (we're all polyglots, and not all image experts) that `@` maps to the `__matmul__` dunder-method

johnkerl (approved):
LGTM with minor suggestions

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,521 @@
...
+        name: Name of the axis.
+        unit: Optional string units. Defaults to ``None``.
+        scale: Optional scale for units. Defaults to ``None``.
+
+    Lifecycle: experimental
+    """
+
+    name: str
+    units: Optional[str] = None
+    scale: Optional[np.float64] = None
```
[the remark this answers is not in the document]
jp-dark: No, because the coordinates live apart from the transforms. Without the scale, it would be implied that the units are always factors of 1. But that's not necessary true.

For example, you can have a coordinate system where a single unit is "10 inches". I think it's nice to split the unit type and the scale of the unit into separate pieces. It also gives me the option of adjusting the scale if I applying a scale transformation to a coordinate space.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,521 @@
...
+        Lifecycle: experimental
+        """
+        return tuple(axis.name for axis in self._axes)
+
+    def rank(self) -> int:
+        """The number of axes in this coordinate space.
+
+        Lifecycle: experimental
+        """
+        return len(self)
```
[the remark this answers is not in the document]
jp-dark: No. I can drop it.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,521 @@
...
+        """Returns the inverse coordinate transform.
+
+        Lifecycle: experimental
+        """
+        return ScaleTransform(
+            self.output_axes, self.input_axes, 1.0 / self._scale_factors
+        )
+
+    @property
+    def isotropic(self) -> bool:
```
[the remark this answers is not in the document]
jp-dark: I debated creating an `UniformScaleTransform` subclass for uniform scale transforms instead of adding it as a property. Thinking through your questions here convinced me that is more correct, since it's really what I'm trying to support anyway.

python-spec/src/somacore/coordinates.py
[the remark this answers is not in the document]
ivirshup: > For example, you can have a coordinate system where a single unit is "10 inches".

I'm not sure I understand the use case for this, could you elaborate? If I did, I think I would define that as one axis whose unit was `"10 inches"` then have a scale transform to another coordinate system that was in inches.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,521 @@
...
+    Args:
+        name: Name of the axis.
+        unit: Optional string units. Defaults to ``None``.
+        scale: Optional scale for units. Defaults to ``None``.
+
+    Lifecycle: experimental
+    """
+
+    name: str
+    units: Optional[str] = None
```
ivirshup, line 25:
```suggestion
    unit: Optional[str] = None
```

Should this be singular, `unit`?

ivirshup: It is singular in the doc string

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,521 @@
...
+                    self.output_axes,
+                    self.augmented_matrix @ other.augmented_matrix,
+                )
+        if isinstance(other, np.ndarray):
+            raise NotImplementedError(
+                "Support for multiplying by numpy arrays is not yet implemented."
+            )
+        raise TypeError(
+            f"Cannot multiply a CoordinateTransform by type {type(other)!r}."
+        )
```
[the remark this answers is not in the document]
jp-dark: The singledispatch didn't play nice with the abcmethods in this case, but I did clean-up the `__matmul__` and `__rmatmul__` methods.

python-spec/src/somacore/__init__.py
```diff
@@ -61,4 +67,10 @@
     "AxisQuery",
     "ExperimentAxisQuery",
     "ContextBase",
+    "Axis",
+    "CoordinateSpace",
+    "CoordinateTransform",
+    "AffineTransform",
+    "ScaleTransform",
+    "IdentityTransform",
```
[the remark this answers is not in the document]
jp-dark: If it is useful. Right now I don't see the value of it over just composing the transform, but I am open to adding it.

python-spec/src/somacore/coordinates.py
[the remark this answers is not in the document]
jp-dark: I renamed them to `unit_name` and `unit_scale`. Does that help clarify?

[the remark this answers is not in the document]
ivirshup: I don't think the names are an issue.

Could you elaborate on use case for this? I'm having trouble understanding why you would want to have this when you can already define scaling with the transformations.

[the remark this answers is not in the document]
jp-dark: The coordinate system exists outside of the transformation. The unit name and scale isn't describing an operation, it's telling you what the fixed size of a single unit for that axis is.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,470 @@
...
+import abc
+import collections.abc
+from dataclasses import dataclass
+from typing import Any, Optional, Sequence, Tuple, Union
+
+import numpy as np
+import numpy.typing as npt
+
+
+@dataclass
```
bkmartinjr, line 12: Historically we have used `attrs` over `dataclasses` in the SOMA API surface (I believe it is only used in a few places at the moment). This was based on the recommendation of @thetorpedodog 

Is there a compelling reason to use `dataclasses` here, vs `attrs`?  My hot take is we should pick one and stick with it across the entire API.

johnkerl: yikes

i totally missed seeing that -- but bruce is quite right

i wish we had a way to autodetect this ... maybe a ci script that greps for '@dataclass' -- ?

bkmartinjr: @jp-dark - I see the same pattern in other PRs (e.g., #219). I won't note those, but ideally we would apply the same (TBD) policy to all of them.  I'm personally quite comfortable with using `attrs`, but happy to chat about it if there is a specific motivation to introduce `dataclasses`

bkmartinjr (approved):
LGTM

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+        self,
+        input_axes: Union[str, Sequence[str]],
+        output_axes: Union[str, Sequence[str]],
+    ):
+        self._input_axes = (
+            (input_axes,) if isinstance(input_axes, str) else tuple(input_axes)
+        )
+        self._output_axes = (
+            (output_axes,) if isinstance(output_axes, str) else tuple(output_axes)
+        )
```
thetorpedodog, line 117: we could probably add a function to do this next to `is_nonstringy_sequence`.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+    def __rmatmul__(self, other: "CoordinateTransform") -> "CoordinateTransform":
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def inverse_transform(self) -> "CoordinateTransform":
+        """Returns the inverse coordinate transform.
+
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
```
thetorpedodog, line 133: Should this one return a `Self`?

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+import numpy.typing as npt
+
+
+class Axis:
+    """A description of an axis of a coordinate system
+
+    Args:
+        name: Name of the axis.
+        unit_name: Optional string name for the units. Defaults to ``None``.
+        unit_scale: Optional scale for units. Defaults to ``None``.
```
thetorpedodog, line 17: When specifying these with `attrs`, you can put the docstring immediately below the field:

```python
@attrs.define(frozen=True)
class Axis:
  """A description of an axis ..."""

  name: str
  """The name of the axis."""
  unit_name: Optional[str] = None
  """Optional string name ..."""
  # etc.
```

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+        """Optional string name for the physical quantity the units are measured in."""
+        return self._unit_name
+
+    @property
+    def unit_scale(self) -> Optional[np.float64]:
+        """Optional scale for the physical quantity the units."""
+        return self._unit_scale
+
+
+class CoordinateSpace(collections.abc.Sequence):
```
thetorpedodog, line 50: This can also be an `attrs` class, with a `validator`:

```
@attrs.define(frozen=True)
class CoordinateSpace(collections.abc.Sequence[Axis]):
  """..."""

  axes: Tuple[Axis, ...] = attrs.field(converter=tuple)

  @axes.validator
  def _validate(self, _, axes: Tuple[Axis, ...]) -> None:
    if not axes:
      raise ValueError("at least one")
    if len(set(...)) != len(axes):
      raise ValueError("unique")
```

This will also give you `__repr__` for free.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+        matrix: npt.ArrayLike,
+    ):
+        super().__init__(input_axes, output_axes)
+
+        # Check the rank of the input/output axes match.
+        if len(self.input_axes) != len(self.output_axes):
+            raise ValueError(
+                "The input axes and output axes must be the same length for an "
+                "affine transform."
+            )
```
thetorpedodog, line 180: It looks like this is true for all types of transform. Should this check be moved to the base `CoordinateTransform`?

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+                    np.hstack((self._matrix, np.zeros((rank, 1)))),
+                    np.hstack((np.zeros((rank,)), np.array([1]))),
+                )
+            )
+        else:
+            raise ValueError(
+                f"Unexpected shape {self._matrix.shape} for the input affine matrix."
+            )
+
+    def __matmul__(self, other: Any) -> CoordinateTransform:
```
thetorpedodog, line 213: `other` should probably be marked as an `object`, with `if not isinstance(object, CoordinateTransform): return NotImplemented` as the first statement.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+            return AffineTransform(other.input_axes, self.output_axes, self._matrix)
+        if isinstance(other, AffineTransform):
+            return AffineTransform(
+                other.input_axes,
+                self.output_axes,
+                self.augmented_matrix @ other.augmented_matrix,
+            )
+        raise TypeError(
+            f"Cannot multiply a CoordinateTransform by type {type(other)!r}."
+        )
```
thetorpedodog, line 226: Instead of `raise TypeError` here, it is better to [`return NotImplemented` so the interpreter can do the right thing](https://docs.python.org/3/reference/datamodel.html#object.__matmul__). I think doing so may allow you to remove the `__rmatmul__` method.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,491 @@
...
+    Lifecycle: experimental
+    """
+
+    def __init__(
+        self,
+        input_axes: Union[str, Sequence[str]],
+        output_axes: Union[str, Sequence[str]],
+        scale_factors: npt.ArrayLike,
+    ):
+        super(AffineTransform, self).__init__(input_axes, output_axes)
```
thetorpedodog, line 288: Would it be better to generate a matrix from the `scale_factors` to pass to the parent? Right now this ends up skipping over the `AffineTransform`’s `__init__`, which means it doesn’t fulfill the expectation that `_matrix` is set which other methods from `AffineTransform` would expect.

A few solutions:

- Pass through the expected inputs to each (so `UniformScaleTransform` calls `super().__init__(ins, outs, [scale_factor] * len(ins)`; `ScaleTransform` calls `super().__init__(ins, outs, scale_factors_matrix)`, etc. That will ensure that all the invariants of the entire class hierarchy are satisfied.

- Use an inheritance structure so all the shared stuff lives in `_AffineBase`, and each subclass contains only the behavior specific to its type:
  ```
  _AffineBase (abstract)
  +-- AffineTransform (has matrix)
  +-- ScaleTransform (has list)
  +-- UniformScaleTransform (has single element)
  ```

- Don’t special-case scalings and uniform scalings, and use a factory method:
  ```python
  class AffineTransform(...):
    def __init__(self, ins, outs, matrix):
      ...

    @classmethod
    def scaling(cls, ins, outs, factors) -> Self:
      ...
      return cls(ins, outs, factor_matrix)

    @classmethod
    def uniform_scaling(cls, ins, outs, factor) -> Self:
      ...
    ```

python-spec/src/somacore/coordinates.py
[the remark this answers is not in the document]
jp-dark: Discussed synchronously. Adding the scale is non-standard and will make syncing with out specs/tools more difficult. If we want to use a coordinate system with non-standard units, we'll just provide the transform as well.

python-spec/src/somacore/coordinates.py
jp-dark: Sweet! That validator is nice -  I can see already why you like `attrs` over `dataclass`.

python-spec/src/somacore/coordinates.py
jp-dark: No, at some point we will want to add in transforms the map between spaces with different number of dimensions. This will be important for 3D support in the long run.

thetorpedodog: Makes sense. It might be worth adding something like

```python
def _check_same_length(self) -> None:
  if len(self.input_axes) != len(self.output_axes):
    raise ValueError(
      f"The input axes and output axes of a {type(self).__name__} must be the same length."
    )
```

so that you can add `self._check_same_length` into only the initializers that need it.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,453 @@
...
+                )
+            )
+        else:
+            raise ValueError(
+                f"Unexpected shape {self._matrix.shape} for the input affine matrix."
+            )
+
+    def __matmul__(self, other: Any) -> CoordinateTransform:
+        if self.input_axes != other.output_axes:
+            raise ValueError("Axis mismatch between transformations.")
```
thetorpedodog, line 177: Probably also worth adding this as something similar to `_check_same_length` above, so that you don’t need to repeat these lines (and you can maybe give the user a slightly better error message, like "output axes of {self_type} must match input axes of {other_type}"

python-spec/src/somacore/types.py
```diff
@@ -22,6 +31,16 @@ def is_nonstringy_sequence(it: object) -> TypeGuard[Sequence]:
     return not isinstance(it, (str, bytes)) and isinstance(it, Sequence)
 
 
+def to_string_tuple(obj: Union[str, Sequence[str]]) -> Tuple[str, ...]:
+    """Returns a tuple of string values.
+
+    If the input is a string, it is returned as a tuple with the sting as its
```
thetorpedodog, line 37: <pre>
>>> to_string_tuple("one")
(Sting, the musician,)
</pre>

```suggestion
    If the input is a string, it is returned as a tuple with the string as its
```

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,445 @@
...
+
+    Lifecycle: experimental
+    """
+
+    axes: Tuple[Axis, ...] = attrs.field(converter=tuple)
+
+    @axes.validator
+    def _validate(self, _, axes: Tuple[Axis, ...]) -> None:
+        if not axes:
+            raise ValueError("at least one")
```
thetorpedodog, line 43: whoops, the strings I used in my example above were just short references to what the full message should be; these will be raised as-is, so they should have the full description.

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,445 @@
...
+    """
+
+    name: str
+    """Name of the axis."""
+    unit: Optional[str] = None
+    """Optional string name for the units of the axis."""
+
+
+@attrs.define(frozen=True)
+class CoordinateSpace(collections.abc.Sequence):
```
thetorpedodog, line 29: `collections.abc.Sequence[Axis]`

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,445 @@
...
+    def _validate(self, _, axes: Tuple[Axis, ...]) -> None:
+        if not axes:
+            raise ValueError("at least one")
+        if len(set(axis.name for axis in self.axes)) != len(axes):
+            raise ValueError("unique")
+
+    def __len__(self) -> int:
+        return len(self.axes)
+
+    def __getitem__(self, index: int) -> Axis:  # type: ignore[override]
```
thetorpedodog, line 50: doing the `Sequence[Axis]` change above means this `type: ignore` should be able to go away.

python-spec/src/somacore/coordinates.py
thetorpedodog: bump on the idea above, to reduce the duplication in these `__init__`s (and make the errors nice and consistent)

python-spec/src/somacore/coordinates.py
thetorpedodog: you missed removing this if-statement!

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,445 @@
...
+            )
+        if isinstance(other, AffineTransform):
+            return AffineTransform(
+                self.input_axes,
+                other.output_axes,
+                other.augmented_matrix @ self.augmented_matrix,
+            )
+        raise NotImplementedError(
+            f"Cannot multiply a CoordinateTransform by type {type(other)!r}."
+        )
```
thetorpedodog, line 230: looking again at the logic here: right now there is a certain amount of duplicated logic between the `__matmul__` and the `__rmatmul__` functions. Would it make sense to formulate the matrix multiplications _entirely_ based on `__matmul__`? That would make the code that is executed a bit more predictable, so that `a @ b` will _always_ be `a.__matmul__(b)`, and `a` only has to worry about all the possible `b`s, but `b` doesn’t need to know anything special about `a`. In most circumstances, `b.__rmatmul__(a)` will never be called anyway, because within our little ecosystem, all `a`s know about all `b`s.

thetorpedodog (commented):
The solution to the class hierarchy situation here is good; I just have a few more suggestions (sorry) mostly about the operator implementation.

python-spec/src/somacore/coordinates.py
jp-dark: ... I'm embarrassed I didn't catch that when I copied it over.

python-spec/src/somacore/coordinates.py
jp-dark: I intentionally left that inline since it's only called in the one place now that I'm using proper inheritance. Do you think it should still create a new function anyway?

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,445 @@
...
+
+    @property
+    def augmented_matrix(self) -> npt.NDArray[np.float64]:
+        """Returns the augmented affine matrix for the transformation.
+
+        Lifecycle: experimental
+        """
+        return self._matrix
+
+    def inverse_transform(self) -> CoordinateTransform:
```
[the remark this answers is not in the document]
jp-dark: Nice.

johnkerl (approved):
🚢

aaronwolen (approved):

jp-dark:
[sc-50289]

python-spec/src/somacore/coordinates.py
```diff
@@ -0,0 +1,381 @@
...
+    """
+
+    name: str
+    """Name of the axis."""
+    unit: Optional[str] = None
+    """Optional string name for the units of the axis."""
+
+
+@attrs.define(frozen=True)
+class CoordinateSpace(
```
nguyenv, line 29: Can we also implement `__eq__` for `CoordinateSpace` in here?

thetorpedodog: coming back from the dead to say that attrs gives you `__eq__` for free

PR #219, opened 2024-09-13
Add new spatial foundational types

jp-dark:
New classes:
* SpatialDataFrame: Abstract base class for spatial data frames (point clouds and geometry/polygon data frames).
    * PointCloud: A spatial data frame for point data
    * GeometryDataFrame: A spatial data for polygon data
* MultiscaleImage: An image class that can contain multiple image levels.
    * ImageProperties: A protocol with the required properties for image levels.
* SpatialRead: New dataclass for returning read data with coordinate space information.

Note: This PR is ready for review, but will be merged after PR #218


python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,623 @@
...
+        domain: Optional[Sequence[Optional[Tuple[Any, Any]]]] = None,
+        platform_config: Optional[options.PlatformConfig] = None,
+        context: Optional[Any] = None,
+    ) -> Self:
+        """Creates a new ``GeometryDataFrame`` at the given URI.
+
+        The schema of the created geoemetry dataframe will include a column named
+        ``soma_joinid`` of type ``pyarrow.int64``, with negative values
+        disallowed, and a column named ``soma_geometry of type ``pyarrow.binary`` or
+        ``pyarrow.large_binary``.  If a ``soma_joinid`` column or ``soma_geometry``
```
XanthosXanthopoulos, line 291: `pyarrow.large_binary` is probably unnecessary to support since a >2GB binary WKB blob is improbable.

bkmartinjr (commented):
Is there a draft set of changes to the abstract spec detailing the new foundational types (e.g., PointCloud)?

python-spec/src/somacore/spatialdata.py
bkmartinjr: it is not uncommon for implementations to promote a `binary` to `large_binary`, so we definitely need to support it.  Concrete use case: code that is doing a "copy" operation (ie.. creating a duplicate of another object) will read back the promoted type, and call create with that value.

bkmartinjr (approved):
No concerns with code, but would very much like to see a companion PR that includes a draft revision of the design (aka abstract) spec.

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,623 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def create(
+        cls,
+        uri: str,
+        *,
+        schema: pa.Schema,
+        index_column_names: Sequence[str] = (options.SOMA_JOINID, "x", "y"),
+        axis_names: Sequence[str] = ("x", "y"),
```
nguyenv, line 217: How do we determine or set which dimensions are axis columns in the TileDB Array? As a metadata field?

jp-dark: Yes. They need to match the axes in the `CoordinateSpace`, so I am effectively storing them with the coordinate space metadata in the TileDB prototype.

jp-dark:
> No concerns with code, but would very much like to see a companion PR that includes a draft revision of the design (aka abstract) spec.

I added a new (WIP) PR #223 here. It should be moving out of draft form soon.

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,623 @@
...
+@dataclass
+class SpatialRead(Generic[_ReadData]):
+    """Reader for spatial data.
+
+    Args:
+        data: The data accessor.
+        data_coordinate_space: The coordinate space the read data is defined on.
+        output_coordinate_space: The requested output coordinate space.
+        coordinate_transform: A coordinate transform from the data coordiante space to
+            the desired output coordiante space.
```
nguyenv, line 597:
```suggestion
        coordinate_transform: A coordinate transform from the data coordinate space to
            the desired output coordinate space.
```

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,623 @@
...
+            region: The region to query. May be a box in the form
+                [x_min, y_min, x_max, y_max] (for 2D images), a box in the form
+                [x_min, y_min, z_min, x_max, y_max, z_max] (for 3D images), or
+                a shapely Geometry.
+            column_names: the named columns to read and return.
+                Defaults to ``None``, meaning no constraint -- all column names.
+            extra_coords: a name to coordinate mapping non-spatial index columns.
+                Defaults to selecting entire region for non-spatial coordinates.
+            transform: An optional coordinate transform that provides desribes the
+                Defaults to ``None``, meaning an identity transform.
```
nguyenv, line 114: Unfinished doc?

johnkerl (approved):

bkmartinjr (approved):

johnkerl (approved):

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+                schema, and at least one index column name is required.
+            axis_names: An ordered list of axis column names that correspond to the
+                names of axes of the the coordinate space the points are defined on.
+                Must be the name of index columns.
+            domain: An optional sequence of tuples specifying the domain of each
+                index column. Each tuple should be a pair consisting of the minimum
+                and maximum values storable in the index column. If omitted entirely,
+                or if ``None`` in a given dimension, the corresponding index-column
+                domain will use the minimum and maximum possible values for the
+                column's datatype.  This makes a point cloud dataframe growable.
```
ivirshup, line 86: My preference would be that it's infinitely growable unless a domain is passed.

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+
+        Args:
+            region: The region to query. May be a box in the form
+                [x_min, y_min, x_max, y_max] (for 2D images), a box in the form
+                [x_min, y_min, z_min, x_max, y_max, z_max] (for 3D images), or
+                a shapely Geometry.
+            column_names: The named columns to read and return.
+                Defaults to ``None``, meaning no constraint -- all column names.
+            extra_coords: A name to coordinate mapping for non-spatial index columns.
+                Defaults to selecting the entire region for non-spatial coordinates.
```
ivirshup, line 378: I'm not sure I understand what this is for

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def read_region(
+        self,
+        region: Optional[options.SpatialRegion] = None,
+        column_names: Optional[Sequence[str]] = None,
+        *,
+        extra_coords: Optional[Mapping[str, options.SparseDFCoord]] = None,
+        transform: Optional[coordinates.CoordinateTransform] = None,
```
ivirshup, line 358: What do we gain by having the transform here instead of the user transforming the result themselves?

Why have a transform argument here, but not for write?

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def create(
+        cls,
+        uri: str,
+        *,
+        type: pa.DataType,
+        image_type: str = "CYX",
+        reference_level_shape: Sequence[int],
```
ivirshup, line 505: I'm still not totally sure on this one. Why is it needed here, but not for spatialdata?

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+
+    # Lifecycle
+
+    @classmethod
+    @abc.abstractmethod
+    def create(
+        cls,
+        uri: str,
+        *,
+        type: pa.DataType,
```
ivirshup, line 503: Does this have to be called `type`? `data_type` or `dtype` could be more appropriate.

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+        image_type: str = "CYX",
+        reference_level_shape: Sequence[int],
+        axis_names: Sequence[str] = ("c", "x", "y"),
```
ivirshup, line 506: Is there any assumed connection between the image_type and the order of the axis names?

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+        result_order: options.ResultOrderStr = _RO_AUTO,
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> "SpatialRead[Union[pa.Tensor, pa.Table]]":
+        """Reads a user-defined region into a :class:`SpatialRead` with data in
+        either an Arrow tensor or table.
+
+        Reads the bounding box of the input region from the requested image level. If
+        ``create_mask=True``, this will return a :class:`SpatialRead` with the image
+        data and mask stored in a :class`pa.Table`. Otherwise, the data will be stored
+        as a :class:`pa.Tensor`.
```
ivirshup, line 567: What is the format of that table? Also what is the size of the returned image if `create_mask=True`?

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def get_transformation_to_level(
+        self, level: Union[int, str]
+    ) -> coordinates.ScaleTransform:
+        """Returns the transformation from the image reference level to the user
+        requested level.
```
ivirshup, line 639: Does the user specify this transformation? I know that there can be multiple algorithms for downscaling, but I forget how this is handled in the new ome transformations.

ivirshup (commented):
Should `SpatialDataFrame` and  `PointCloud` have a `coordinate_space` property? `MultiscaleImage` has this.

To me, it's more important for the "coordinate" types since they can have units other than pixels.

python-spec/src/somacore/spatialdata.py
ivirshup: Update: this was a misunderstanding of what this argument does. The transform is meant for the region.

Maybe will be renamed to something like "`region_transform`"

python-spec/src/somacore/spatialdata.py
```diff
@@ -0,0 +1,763 @@
...
+
+    def __post_init__(self):
+        if (
+            self.data_coordinate_space.axis_names
+            != self.coordinate_transform.input_axes
+        ):
+            raise ValueError(
+                "Input coordinate transform axis names do not match the data coordinate"
+                "space."
+            )
```
nguyenv, line 755: Error message shows coordinatespace without space.

```suggestion
            raise ValueError(
                "Input coordinate transform axis names do not match the data coordinate "
                "space."
            )
```

python-spec/src/somacore/spatialdata.py
jp-dark: This is to set the default/reference size of the multiscale image. This will usually be the shape of the level 0 image.

python-spec/src/somacore/spatialdata.py
jp-dark: It's what `NDArray` and `DenseNDArray` (and pyarrow) use for the argument, so I think we should leave it for consistency.

python-spec/src/somacore/spatialdata.py
jp-dark: The way I currently implemented it, there is only a connection with the location of channel, but that is really confusing (even I'm confused!). So yes, it should match the order and I'll change the order of the default `axis_names` to match the default `image_type`.

python-spec/src/somacore/spatialdata.py
jp-dark: From the meeting: 
* If supported by arrow, we should return a nullable `pyarrow.Tensor` instead of a `Table`. If that isn't supported, we should return tuple of `pa.Table` and we should add additional docs here about the type/size of the mask.

python-spec/src/somacore/spatialdata.py
jp-dark: As discussed in the meetinig: the transform isn't applied to the image. It's returned to the user to do with as they will. We will need to add functionality to make the transforms useful, but it hasn't been implemented yet.

python-spec/src/somacore/spatialdata.py
jp-dark: The behavior here matches the default behavior for the `DataFrame` class. I would like to keep the behavior between the different classes as similar as possible unless there is a compelling reason to change it for a particular type.

python-spec/src/somacore/spatialdata.py
jp-dark: I went ahead and dropped the mask for now.

python-spec/src/somacore/options.py
```diff
@@ -177,5 +181,9 @@ class ResultOrder(enum.Enum):
     pa.ChunkedArray,
 ]
 """A single coordinate range for one dimension of a sparse ndarray."""
+
 SparseNDCoords = Sequence[SparseNDCoord]
 """A sequence of coordinate ranges for reading sparse ndarrays."""
+
+SpatialRegion = Union[Sequence[int], Sequence[float], shapely.GeometryType]
```
nguyenv, line 188:
```suggestion
SpatialRegion = Union[Sequence[int], Sequence[float], shapely.geometry.base.BaseGeometry]
```


I think this actually needs to be `shapely.geometry.base.BaseGeometry` because `shapely.box` returns a `shapely.Polygon` which is a subclass of `shapely.geometry.base.BaseGeometry` but not `shapely.GeometryType`.

```
def test_point_cloud_read_region_basic(tmp_path):
    uri = tmp_path.as_uri()

    schema = pa.schema([("x", pa.float64()), ("y", pa.float64())])

    with soma.PointCloud.create(uri, schema=schema) as ptc:
        pydict = {
            "soma_joinid": [1, 2, 3, 4, 5],
            "x": [10, 20, 30, 40, 50],
            "y": [4.1, 5.2, 6.3, 7.4, 8.5],
        }
        rb = pa.Table.from_pydict(pydict)
        ptc.write(rb)

    with soma.PointCloud.open(uri, "r") as ptc:
        # Must be a bounding box
        with pytest.raises(ValueError):
            ptc.read_region(region=[])

        read_region = ptc.read_region(region=None)
        print(read_region.data.concat())

        # read_region = ptc.read_region(region=[10, 50, 1, 10])
        # print(read_region.data.concat())

        region = shapely.box(10.0, 50.0, 1.0, 10.0)
        print(type(region)) # <class 'shapely.geometry.polygon.Polygon'>
        print(isinstance(region, shapely.Polygon)) # True
        print(isinstance(region, shapely.GeometryType)) # False
        print(issubclass(shapely.Polygon, shapely.GeometryType)) # False
        print(isinstance(region, shapely.geometry.base.BaseGeometry)) # True
        print(issubclass(shapely.Polygon, shapely.geometry.base.BaseGeometry)) # True

        # >       raise TypeCheckError(f"did not match any element in the union:\n{formatted_errors}")
        # E       typeguard.TypeCheckError: argument "region" (shapely.geometry.polygon.Polygon) did not match any element in the union:
        # E         Sequence[int]: is not a sequence
        # E         Sequence[float]: is not a sequence
        # E         shapely._geometry.GeometryType: is not an instance of shapely._geometry.GeometryType
        # E         NoneType: is not an instance of NoneType
        read_region = ptc.read_region(region=region)
```

python-spec/src/somacore/spatialdata.py
jp-dark: We discussed synchronously and agreed to leave this in for now, but to revisit the design during the alpha testing.

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+
+        Returns: a tuple of minimum and maximum values, inclusive,
+            storable on each index column of the dataframe.
+
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
+
+
+class GeometryDataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
```
aaronwolen, line 261: I think we need to align naming of the `PointCloud` and `GeometryDataFrame` classes. Both are essentially specialized `DataFrame`s so the `<specialization>DataFrame` naming scheme makes sense. Although `PointCloudDataFrame` is a bit of a mouthful. I don't have strong feelings about which approach we use but I do think we should be consistent, unless there's a good reason not to.

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+    @abc.abstractmethod
+    def create(
+        cls,
+        uri: str,
+        *,
+        schema: pa.Schema,
+        index_column_names: Sequence[str] = (
+            options.SOMA_JOINID,
+            options.SOMA_GEOMETRY,
+        ),
```
aaronwolen, line 283: Here too I'm not sure if we need to allow users to define the index columns on these specialized dataframes.

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+        axis_names: Sequence[str] = ("x", "y"),
+        domain: Optional[Sequence[Optional[Tuple[Any, Any]]]] = None,
+        platform_config: Optional[options.PlatformConfig] = None,
+        context: Optional[Any] = None,
+    ) -> Self:
+        """Creates a new ``GeometryDataFrame`` at the given URI.
+
+        The schema of the created geoemetry dataframe will include a column named
```
aaronwolen, line 291:
```suggestion
        The schema of the created geometry dataframe will include a column named
```

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+        ``soma_joinid`` of type ``pyarrow.int64``, with negative values
+        disallowed, and a column named ``soma_geometry of type ``pyarrow.binary`` or
+        ``pyarrow.large_binary``.  If a ``soma_joinid`` column or ``soma_geometry``
+        are present in the provided schema, they must be of the correct type.  If
+        either the ``soma_joinid`` column or ``soma_geometry`` column are not provided,
+        one will be added. The ``soma_joinid`` may be an index column. The
+        ``soma_geometry`` column must be an index column.
```
aaronwolen, line 298: If we do allow users to specify the index columns, what happens if they don't pass `"soma_geometry"` to the `index_column_names` arg?

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+                for the particular SOMA implementation for details.
+        Returns:
+            A :class:`ReadIter` of :class:`pa.Table`s.
+
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def read_region(
```
aaronwolen, line 366: Same concern as above. However, since it's possible to query on more complex geometries than just bounding boxes could we update the name to reflect that? Maybe `read_shape()`?

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+    ) -> Self:
+        """Creates a new collection of this type at the given URI.
+
+        Args:
+            uri: The URI where the collection will be created.
+            reference_level_shape: The shape of the reference level for the multiscale
+                image. In most cases, this should correspond to the size of the image
+                at ``level=0``.
+            axis_names: The names of the axes of the image.
+            axis_types: The types of the axes of the image. Must be the same lenght as
```
aaronwolen, line 547:
```suggestion
            axis_types: The types of the axes of the image. Must be the same length as
```

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+        Args:
+            level: The image level to read the data from. May use index of the level
+                or the image name.
+            region: The region to query. May be a box in the form
+                [x_min, y_min, x_max, y_max] (for 2D images), a box in the form
+                [x_min, y_min, z_min, x_max, y_max, z_max] (for 3D images), or
+                a shapely Geometry.
+            channel_coords: An optional slice that defines the channel coordinates
+                to read.
+            region_transform: An optional coordinate transform that provides desribes the
```
aaronwolen, line 606:
```suggestion
            region_transform: An optional coordinate transform that provides the
```

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+                transformation from the provided region to the reference level fo this
```
aaronwolen, line 607:
```suggestion
                transformation from the provided region to the reference level of this
```

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+            result_order: the order to return results, specified as a
+                :class:`~options.ResultOrder` or its string value.
+
+        Returns:
+            The data bounding the requested region as a :class:`SpatialRead` with
+            :class:`pa.Tensor` data.
+        """
+        raise NotImplementedError()
+
+    # Metadata opeations
```
aaronwolen, line 621:
```suggestion
    # Metadata operations
```

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+        raise NotImplementedError()
+
+
+class MultiscaleImage(  # type: ignore[misc]  # __eq__ false positive
+    base.SOMAObject,
+    Generic[_DenseND, _RootSO],
+    MutableMapping[str, _DenseND],
+    metaclass=abc.ABCMeta,
+):
+    """A multiscale image with an extendable number of levels.
```
aaronwolen, line 502:
```suggestion
    """A multiscale image with an extendable number of resolution levels.
```

Trying to make it clearer to users what a "level" is in this context.

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+
+    The multiscale image defines the top level properties. Each level must
+    match the expected following properties:
```
aaronwolen, line 505: Seems like some kind of list was intended to follow this?

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+        axis_types: Sequence[str] = ("channel", "height", "width"),
+        platform_config: Optional[options.PlatformConfig] = None,
+        context: Optional[Any] = None,
+    ) -> Self:
+        """Creates a new collection of this type at the given URI.
+
+        Args:
+            uri: The URI where the collection will be created.
+            reference_level_shape: The shape of the reference level for the multiscale
+                image. In most cases, this should correspond to the size of the image
```
aaronwolen, line 544:
```suggestion
                image. In most cases, this corresponds to the size of the image
```

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+    @abc.abstractmethod
+    def coordinate_space(self, value: coordinates.CoordinateSpace) -> None:
+        """Coordinate space for this multiscale image.
+
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def get_transformation_from_level(
```
aaronwolen, line 651: Could we replace `transformation` to `transform`?

aaronwolen (changes requested):
Looks great! I have a few specific comments, questions, and suggestions.

Also wondering if we could/should hold off on adding the `GeometryDataFrame` class? Unless I'm mistaken, I don't believe it's needed for the spot-based assays we're targeting for the initial release? If we do leave it in, I think the docstrings for `PointCloud` and `GeometryDataFrame` should make it clearer how they differ from `SOMADataFrame` and from each other.

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+    soma_type: Final = "SOMAPointCloud"  # type: ignore[misc]
+
+    @classmethod
+    @abc.abstractmethod
+    def create(
+        cls,
+        uri: str,
+        *,
+        schema: pa.Schema,
+        index_column_names: Sequence[str] = (options.SOMA_JOINID, "x", "y"),
```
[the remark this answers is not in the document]
jp-dark: The spatial axes need to be index columns (see doc string), but this is still an argument so there is flexibility with:
1. The order of the index columns.
2. The option to include or exclude `soma_joinid` (or other columns) as index columns.

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+                for the particular SOMA implementation for details.
+        Returns:
+            A :class:`ReadIter` of :class:`pa.Table`s.
+
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def read_region(
```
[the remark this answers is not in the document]
jp-dark: This can be a bounding box or geometry query. I originally had `spatial_read`, but I felt like that was vague. How about `read_spatial_region`?

python-spec/src/somacore/spatial.py
```diff
@@ -0,0 +1,774 @@
...
+    def index_column_names(self) -> Tuple[str, ...]:
+        """The names of the index (dimension) columns.
+
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
+
+    @property
+    @abc.abstractmethod
+    def coordinate_space(self) -> Optional[coordinates.CoordinateSpace]:
```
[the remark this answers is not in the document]
jp-dark: It's generated from the axis names. We could switch the parameter to coordinate space instead. How would you feel about leaving that as a question for alpha users?

python-spec/src/somacore/spatial.py
jp-dark: An error is thrown.

python-spec/src/somacore/spatial.py
jp-dark: Same as above. The user needs to include the geometry as an index column. An error is thrown if they do not. But there is flexibility in the other columns they may want to include.

aaronwolen (approved):
🚀

PR #220, opened 2024-09-13
Add the `Scene` class

jp-dark:
* Create the `Scene` class
* Add `spatial` collection of scenes to the `Experiment` class
* Add Scene to ephemeral collections and tests

This PR is ready for review, but should be merged after PR #219


bkmartinjr (approved):

johnkerl (approved):

jp-dark:
[sc-55577], [sc-55578], [sc-55579]

python-spec/src/somacore/scene.py
```diff
@@ -0,0 +1,270 @@
...
+    @coordinate_space.setter
+    @abc.abstractmethod
+    def coordinate_space(self, value: coordinates.CoordinateSpace) -> None:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def register_geometry_dataframe(
+        self,
+        key: str,
+        transform: coordinates.CoordinateTransform,
```
ivirshup, line 107: Why is there a transform here? I thought everything in the `Scene` had to be in the same coordinate space?

python-spec/src/somacore/scene.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        raise NotImplementedError()
+
+    @coordinate_space.setter
+    @abc.abstractmethod
+    def coordinate_space(self, value: coordinates.CoordinateSpace) -> None:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def register_geometry_dataframe(
```
ivirshup, line 104: Soma collection objects like `Experiment` and `Measurement` have methods like `add_new_dataframe`. Why does this class have `register_` methods instead?

python-spec/src/somacore/scene.py
```diff
@@ -0,0 +1,270 @@
...
+        a point cloud stored in the scene.
+
+        If the subcollection the geometry dataframe is inside of is more than one
+        layer deep, the input should be provided as a sequence of names. For example,
+        to register a geometry dataframe named  "transcripts" in the "var/RNA"
+        collection::
+
+            scene.register_geometry_dataframe(
+                'transcripts', transform, subcollection=['var', 'RNA'],
+            )
```
ivirshup, line 122: Is this a workflow that is being supported? I thought the spec was pretty opinionated about where each kind of thing would be stored.

ivirshup (commented):
I think this will need another pass from me, but I left comments about some of the more high level parts of this PR I'm not sure if I understand.

python-spec/src/somacore/scene.py
jp-dark: Everything needs to be mappable to the same coordinate space.

python-spec/src/somacore/scene.py
jp-dark: This has those methods as well (they are inherited from the `BaseCollection`). These methods record a transform from the `Scene` to the assets.

python-spec/src/somacore/scene.py
jp-dark: The example in the docstring follows the storage guideline and is more than one layer deep.

python-spec/src/somacore/scene.py
aaronwolen: Could we name these methods using the same scheme used for the getters: `<op>_transformation_<class>()`?
```suggestion
    def set_transformation_geometry_dataframe(
```

python-spec/src/somacore/scene.py
```diff
@@ -0,0 +1,270 @@
...
+
+        Returns:
+            Coordinate transform from the scene to the requested dataframe.
+
+        Lifecycle: experimental
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def get_transformation_to_multiscale_image(
```
aaronwolen, line 228:
```suggestion
    def get_transformation_multiscale_image(
```

Do we need `to` here?

aaronwolen (approved):
Had a few naming suggestions but otherwise looks great.

python-spec/src/somacore/scene.py
jp-dark: We need something to specify what the transform is going from/to. (Scene to image is different than image to scene). I don't love how verbose the methods are, so I'm open to other suggestions. I did change `transformation` to match `transform` (based on one of your suggestion changes on #219).

PR #221, opened 2024-09-13
Add spatial presence matrices

jp-dark:
This adds dataframes for creating a join table between the scenes and observations/variables. New methods are added to the `ExperimentAxisQuery` class for getting a Arrow array of scene names that relate to the obs or var in the query.

This PR is ready to be reviewed, but should be merged after PR #220

### DIFF
--- a/python-spec/src/somacore/experiment.py
+++ b/python-spec/src/somacore/experiment.py
@@ -66,6 +66,17 @@ class Experiment(
     spatial = _mixin.item[_SceneColl]()  # TODO: Discuss the name of this element.
     """A collection of named spatial scenes."""
 
+    obs_spatial_presence = _mixin.item[_DF]()
+    """A dataframe that stores the presence of obs in the spatial scenes.
+
+    This provides a join table for the obs ``soma_joinid`` and the scene names used in
+    the ``spatial`` collection. This dataframe must contain dimensions ``soma_joinid``
+    and ``scene_id``. The ``scene_id`` dimension must be a ``string`` dimension. The
+    dataframe must contain a ``boolean`` column ``data``. The values of ``data`` are
+    ``True`` if the obs with obsid ``soma_joinid`` is contained in the scene
+    ``scene_id`` and ``False`` otherwise.
+    """
+
     def axis_query(
         self,
         measurement_name: str,
--- a/python-spec/src/somacore/experiment.py
+++ b/python-spec/src/somacore/experiment.py
@@ -72,8 +72,8 @@ class Experiment(
     This provides a join table for the obs ``soma_joinid`` and the scene names used in
     the ``spatial`` collection. This dataframe must contain index columns ``soma_joinid``
     and ``scene_id``. The ``scene_id`` column must have type ``string``. The
-    dataframe must contain a ``boolean`` column ``data``. The values of ``data`` are
-    ``True`` if the obs with obsid ``soma_joinid`` is contained in the scene
+    dataframe must contain a ``boolean`` column ``soma_data``. The values of ``soma_data`` are
+    ``True`` if the obs ``soma_joinid`` is contained in the scene
     ``scene_id`` and ``False`` otherwise.
     """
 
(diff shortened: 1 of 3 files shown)


python-spec/src/somacore/experiment.py
```diff
@@ -66,6 +66,17 @@ class Experiment(
     spatial = _mixin.item[_SceneColl]()  # TODO: Discuss the name of this element.
     """A collection of named spatial scenes."""
 
+    obs_spatial_presence = _mixin.item[_DF]()
+    """A dataframe that stores the presence of obs in the spatial scenes.
+
+    This provides a join table for the obs ``soma_joinid`` and the scene names used in
+    the ``spatial`` collection. This dataframe must contain dimensions ``soma_joinid``
```
bkmartinjr, line 73: the use of the word `dimension` is confusing as used here.  SOMA does not have the concept of a dimension as you mean it (ie, an index), outside of joinids.

Two suggestions:
1. The abstract spec needs to be updated to clarify the new concepts and changes implied by SpatialDataFrame, et al.
2. We should nail down our terminology and use it consistent.  The abstract spec mentions "index" and "indexing", but does not use the TileDB-specific "dimension" terminology.

python-spec/src/somacore/measurement.py
```diff
@@ -99,3 +99,14 @@ class Measurement(
 
     This is indexed by ``[varid_1, varid_2]``.
     """
+
+    var_spatial_presence = _mixin.item[_DF]()
+    """A dataframe that stores the presence of var in the spatial scenes.
+
+    This provides a join tabel for the var ``soma_joinid`` and the scene names used in
+    the ``spatial`` collection. This dataframe must contain dimensions ``soma_joinid``
+    and ``scene_id``. The ``scene_id`` dimension must be a ``string`` dimension. The
```
bkmartinjr, line 108: `s/dimension/index/`

bkmartinjr (approved):
A couple of minor nits, but otherwise :+1:

python-spec/src/somacore/experiment.py
johnkerl: +1

johnkerl (approved):

ivirshup (approved):
lgtm!

python-spec/src/somacore/measurement.py
```diff
@@ -99,3 +99,14 @@ class Measurement(
 
     This is indexed by ``[varid_1, varid_2]``.
     """
+
+    var_spatial_presence = _mixin.item[_DF]()
+    """A dataframe that stores the presence of var in the spatial scenes.
+
+    This provides a join tabel for the var ``soma_joinid`` and the scene names used in
```
aaronwolen, line 106:
```suggestion
    This provides a join table for the var ``soma_joinid`` and the scene names used in
```

python-spec/src/somacore/experiment.py
```diff
@@ -66,6 +66,17 @@ class Experiment(
     spatial = _mixin.item[_SceneColl]()  # TODO: Discuss the name of this element.
     """A collection of named spatial scenes."""
 
+    obs_spatial_presence = _mixin.item[_DF]()
+    """A dataframe that stores the presence of obs in the spatial scenes.
+
+    This provides a join table for the obs ``soma_joinid`` and the scene names used in
+    the ``spatial`` collection. This dataframe must contain index columns ``soma_joinid``
+    and ``scene_id``. The ``scene_id`` column must have type ``string``. The
+    dataframe must contain a ``boolean`` column ``data``. The values of ``data`` are
```
aaronwolen, line 75:
```suggestion
    dataframe must contain a ``boolean`` column ``soma_data``. The values of ``soma_data`` are
```

To make it consistent with reserved columns in other soma cases.

python-spec/src/somacore/experiment.py
```diff
@@ -66,6 +66,17 @@ class Experiment(
...
+    ``True`` if the obs with obsid ``soma_joinid`` is contained in the scene
```
aaronwolen, line 76:
```suggestion
    ``True`` if the obs ``soma_joinid`` is contained in the scene
```

aaronwolen (approved):
A couple minor comments but still good to go.

PR #186, opened 2024-02-13
Support Python 3.12

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/issues/1849

Note: I ran `python-spec/update-requirements-txt`

PR #222, opened 2024-09-16
Work around various issues with Python 3.12

johnkerl:
TL;DR we are trying to support Python 3.12 in [TileDB-SOMA](https://github.com/single-cell-data/TileDB-SOMA)

Solved challenge:

As tracked from parent issue https://github.com/single-cell-data/TileDB-SOMA/issues/1849

* https://github.com/single-cell-data/TileDB-SOMA/pull/2692
* https://github.com/single-cell-data/TileDB-SOMA/pull/2734
* https://github.com/TileDB-Inc/TileDB/pull/5223 went into [core 2.26](https://github.com/TileDB-Inc/TileDB/releases/tag/2.26.0)
* Now it's possible to put up PR https://github.com/single-cell-data/TileDB-SOMA/pull/3001, and hope to release that in TileDB-SOMA 1.14 (https://github.com/single-cell-data/TileDB-SOMA/issues/2999)

Current unsolved challenges:

* Latest Python 3.12 is 3.12.5
  * As debug I tried Python 3.12.3 which worked as debug, but isn't a good option for shipping
* TileDB-SOMA's current `typeguard` version is 4.2.0
* Python 3.12.4 added a new required argument to `evaluate_forward_ref` within Python's built-in `typing` module which broke many packages including `typeguard`
* As a result, our TileDB-SOMA CI for [PR 3001](https://github.com/single-cell-data/TileDB-SOMA/pull/3001) fails [like this](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/10892424953/job/30225213572#step:13:17210)
  * `TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'`
* `typeguard` 4.3.0 contains a fix for this problem
  * https://github.com/agronholm/typeguard/releases
* But this surfaces another problem which is:
```
E       typeguard.TypeCheckError: item 0 of argument "coords" (tuple) did not match any element in the union:
E         NoneType: is not an instance of NoneType
E         int: is not an instance of int
E         somacore.types.Slice[int]: is not compatible with the Slice protocol because it has no attribute named '__orig_bases__'

/home/ubuntu/.pyenv/versions/3.12.5/lib/python3.12/site-packages/typeguard/_checkers.py:433: TypeCheckError
```

The current PR is an attempt to deal with this.

Once this works, we'll need to:

* Merge this PR
* Tag a release of this repo (`somacore`)
* Depend on 1.0.next of `somacore` within the TileDB-SOMA repo

I've currently got the following setup on an EC2 instance:

* `somacore` locally with `pip install -e` with this PR checked out
* `tiledbsoma` locally with `pip install -e` with https://github.com/single-cell-data/TileDB-SOMA/pull/3001 checked out


python-spec/src/somacore/types.py
```diff
@@ -57,6 +57,16 @@ def stop(self) -> Optional[_T_co]: ...
...
 
+    if sys.version_info >= (3, 12):
+
+        @property
+        def __orig_bases__(self) -> Any:
+            # Only exists in Python 3.12 and above
+            from types import get_original_bases
+
+            return get_original_bases(int)
+            # return get_original_bases(_T_co)
```
thetorpedodog, line 68: This is a typeguard bug—it’s checking too many built-in attributes of the protocol that don’t actually matter. (See, for example, [a case where it rejects objects with a completely empty protocol](https://github.com/agronholm/typeguard/issues/479#issuecomment-2356689989) that I found.

That said, if implementing this makes typeguard stop complaining for now, I would do two things:

1. Add a comment to that effect, and that it should be removed at some point
2. Make this function instead raise `AssertionError`, since it should never be called.

ryan-williams:
I believe this PR is no longer necessary; tests+lint have passed on 3.12 in this repo since #186, and https://github.com/single-cell-data/TileDB-SOMA/pull/3001 has 3.12 passing without requiring any somacore changes.

johnkerl:
Awesome, thanks @ryan-williams !

PR #224, opened 2024-09-23
Drop Python 3.8 support

ryan-williams:
Python 3.8 EOL [is 2024-10](https://devguide.python.org/versions/)

PR #225, opened 2024-09-23
CI version bumps, deps factor

ryan-williams:
`requirements-py3.8-lint.txt` repeated all the deps from `requirements-py3.8.txt`; here I've removed those common deps from the former, and install/cache based on both files in GHA. This was motivated by dropping Python 3.8 (#224), where I wanted to audit (and upgrade pins for) the lint-only deps.

Also bumped:
- `actions/checkout`: `v3` → `v4`
- `actions/setup-python`: `v4` → `v5`


.github/workflows/python-somacore.yaml
```diff
@@ -20,11 +20,11 @@ jobs:
...
           python-version: ${{ env.PYTHON_VERSION }}
           cache: pip
-          cache-dependency-path: python-spec/requirements-py${{ env.PYTHON_VERSION }}-lint.txt
+          cache-dependency-path: "python-spec/requirements-py${{ env.PYTHON_VERSION }}*.txt"
 
       - name: Install static analysis packages
-        run: |
-          pip install -r python-spec/requirements-py${{ env.PYTHON_VERSION }}-lint.txt
+        working-directory: python-spec
+        run: pip install -r requirements-py${PYTHON_VERSION}.txt -r requirements-py${PYTHON_VERSION}-lint.txt
```
ryan-williams, line 27: #224 also makes this change, either one can be merged first

johnkerl (approved):

johnkerl:
See also https://github.com/single-cell-data/TileDB-SOMA/issues/2141

PR #229, opened 2024-09-25
Restore access to filtered axism arrays from a query

aaronwolen:
This addresses a bug introduced in #216 that caused obsm/varm exports from `query` objects to ignore filtered joinids.


johnkerl (approved):

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/issues/3075

PR #230, opened 2024-09-26
Expand spatial dataframe docstrings

aaronwolen:
As discussed in https://github.com/single-cell-data/SOMA/pull/219#pullrequestreview-2329701066.

### DIFF
--- a/python-spec/src/somacore/spatial.py
+++ b/python-spec/src/somacore/spatial.py
@@ -265,7 +265,7 @@ class PointCloud(base.SOMAObject, metaclass=abc.ABCMeta):
 class GeometryDataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
     """A specialized SOMA object for storing complex geometries with spatial indexing.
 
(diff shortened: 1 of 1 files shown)


python-spec/src/somacore/spatial.py
```diff
@@ -259,8 +263,10 @@ def domain(self) -> Tuple[Tuple[Any, Any], ...]:
 
 
 class GeometryDataFrame(base.SOMAObject, metaclass=abc.ABCMeta):
-    """A multi-column table of geometries with spatial indexing and a user-defined
-    schema.
+    """A specialized SOMA object for storing complex geometries with spatial indexing.
+
+    The `GeometryDataFrame` class is designed to store and manage geometric shapes such as
```
johnkerl, line 268: double backticks for readthedocs please

aaronwolen: Thanks!

jp-dark (approved):

PR #238, opened 2024-10-22
Update the MultiscaleImage API

jp-dark:
This PR addresses feedback from PR #219 regarding the `MultiscaleImage` API. It includes the following changes:

* Directly take the `CoordinateSpace` as a creation parameter.
* Remove confusing `image_type` property/creation parameter. Use `data_axis_order` instead which uses the axis names the user provided.
* Create the first resolution level when creating the `MultiscaleImage`.
* Require the `add_new_level` to only add images smaller than the base (level=0) image.
* Add a `set` method for adding images that exist outside of SOMA.
* Add a property to check the number of channels in the image.
* Remove `ImageProperties` class.

### DIFF
--- a/python-spec/src/somacore/spatial.py
+++ b/python-spec/src/somacore/spatial.py
@@ -589,6 +589,37 @@ class MultiscaleImage(  # type: ignore[misc]  # __eq__ false positive
         """
         raise NotImplementedError()
 
+    @abc.abstractmethod
+    def set(
+        self,
+        key: str,
+        value: data.DenseNDArray,
+        *,
+        use_relative_uri: Optional[bool] = None,
+    ):
+        """Sets a new level in the multi-scale image to be an existing SOMA
+        :class:`data.DenseNDArray`.
+
+        Args:
+            key: The string key to set.
+            value: The SOMA object to insert into the collection.
+            use_relative_uri: Determines whether to store the collection
+                entry with a relative URI (provided the storage engine
+                supports it).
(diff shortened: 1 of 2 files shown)


python-spec/src/somacore/spatial.py
```diff
@@ -685,54 +673,13 @@ def level_count(self) -> int:
...
-
-        Lifecycle: experimental
-        """
-
-    @property
-    def shape(self) -> Tuple[int, ...]:
-        """Size of each axis of the image.
-
-        Lifecycle: experimental
-        """
```
jp-dark, line 734: This class didn't add much value as is. We may want to expand it to include more information or just remove it altogether.

python-spec/src/somacore/spatial.py
```diff
@@ -528,23 +528,29 @@ def create(
...
         *,
         type: pa.DataType,
-        reference_level_shape: Sequence[int],
-        axis_names: Sequence[str] = ("c", "y", "x"),
+        shape: Sequence[int],
+        coordinate_space: Union[Sequence[str], coordinates.CoordinateSpace] = (
+            "x",
+            "y",
+        ),
         axis_types: Sequence[str] = ("channel", "height", "width"),
```
jp-dark, line 536: Is there a better name or better argument to take to distinguish between the on disk data order and the coordinate space?

aaronwolen (approved):
Changes look good!

PR #243, opened 2024-11-01
Specialize `ExperimentAxisQuery.{obs,var}` to `pa.IntegerArray`

ryan-williams:
### Issue

- [`SparseNDArray.read`] requires a [`SparseNDCoords`] arg. This can be a tuple of `pa.IntegerArray`s, but not a tuple of `pa.Array`s.
- We nevertheless pass `pa.Array`s in two places:
  - `ExperimentAxisQuery.X` [passes][`x_layer.read` call] `self._joinids.{obs,var}`; both are `pa.Array`s.
  - `_fast_csr.read_csr` [passes][csr matrix.read call] `{obs,var}_joinids`; both are `pa.Array`s.
- In reality, the values are actually `pa.IntegerArray`s, so it's valid at runtime, but the types don't reflect that.
- IntelliJ was warning me about it; not sure why mypy wasn't catching it.

### Fix

To address this, [`ExperimentAxisQuery.{obs_var}_joinids`] (and some related fields, e.g. [`_JoinIDCache._cached_{obs,var}`]) should be `pa.IntegerArray`s, instead of `pa.Array`s. Narrowing those return types is backwards-compatible, afaict.

[csr call]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/query/query.py#L469-L474
[joinids]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/query/query.py#L169-L181

[`x_layer.read` call]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/query/query.py#L237-L243
[`SparseNDCoords`]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/options.py#L185
[`SparseNDArray.read`]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/data.py#L403-L405

[csr matrix.read call]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/query/_fast_csr.py#L35
[`_JoinIDCache._cached_{obs,var}`]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/query/query.py#L754-L755
[`ExperimentAxisQuery.{obs_var}_joinids`]: https://github.com/single-cell-data/SOMA/blob/python-v1.0.23/python-spec/src/somacore/query/query.py#L169-L181


johnkerl (approved):
LGTM!

bkmartinjr (approved):
LGTM

PR #244, opened 2024-11-07
Make some Experiment/Query classes abstract

ryan-williams:
Move several classes/files to [TileDB-SOMA] (https://github.com/single-cell-data/TileDB-SOMA/pull/3307), leave abstract interfaces here.

[TileDB-SOMA]: https://github.com/single-cell-data/TileDB-SOMA


johnkerl (approved):
LGTM!

ryan-williams:
I pushed one more commit, explicitly tagging `ExperimentAxisQuery`/`AxisIndexer` as `ABC`s.

@jp-dark had suggested using `Protocol`s last week; I'm not clear on what the advantages might be of that.

johnkerl:
> @jp-dark had suggested using Protocols last week; I'm not clear on what the advantages might be of that.

@ryan-williams I think the suggestion has merit but I'd prefer to take it as follow-on work and not let it block this current PR.

PR #249, opened 2024-11-18
Add `coordinate_space` to `Scene.create`

jp-dark:
Allow user to specify the scene coordinate space on creation.

### DIFF
--- a/python-spec/src/somacore/scene.py
+++ b/python-spec/src/somacore/scene.py
@@ -1,14 +1,15 @@
 """Implementation of the SOMA scene collection for spatial data"""
 
 import abc
-from typing import Generic, Optional, Sequence, TypeVar, Union
+from typing import Any, Generic, Optional, Sequence, TypeVar, Union
 
-from typing_extensions import Final
+from typing_extensions import Final, Self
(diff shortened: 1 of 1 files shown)


python-spec/src/somacore/scene.py
```diff
@@ -90,6 +91,35 @@ class Scene(
...
+        ] = None,
+        platform_config: Optional[options.PlatformConfig] = None,
+        context: Optional[Any] = None,
+    ) -> Self:
+        """Creates a new collection of this type at the given URI.
+
+        Args:
+            uri: The URI where the collection will be created.
+            coordinate_space: Optional coordinate space or the axis names for the
+                coordinate space the scene is defined on. Defaults to ``None``.
```
johnkerl, line 111: Say what happens if `None` -- does it mean the resulting `MultiscaleImage` will not have any coordinate space? Or that we will use a default? And if the latter, please say what we will use.

johnkerl (changes requested):

aaronwolen (approved):

johnkerl (approved):

PR #252, opened 2024-12-19
Convert `Optional[]` to union types

mojaveazure:
Convert `Optional[]` to [union types](https://docs.python.org/3/library/stdtypes.html#union-type) in type hints

[SC-61053](https://app.shortcut.com/tiledb-inc/story/61053)

resolves single-cell-data/TileDB-SOMA#3334


johnkerl (approved):
Thank you @mojaveazure !! 🚢

